PR #10, opened 2018-08-28
Adds prettier as a pre-commit hook

### LINKED ISSUE #5: chore: Prettier in git commit hook
We should have a githook that runs prettier so that all code committed is formatted.

I set it up in https://github.com/reach/router with `pretty-quick`

Would love somebody to tackle this :)

sudarshang:
Fixes #5


ryanflorence:
🎉

PR #11, opened 2018-08-28
Remove package-lock.json

Andarist:


ryanflorence:
I think Lerna does this? I haven't followed the world of package-lock v. yarn.lock, is there only supposed to be one? Can people contribute if they don't use yarn? Would love to know.

Andarist:
> I think Lerna does this?

Might be.

> I haven't followed the world of package-lock v. yarn.lock, is there only supposed to be one?

Yeah, there should be only one. yarn i.e. warns u if it sees package-lock.json too.

> Can people contribute if they don't use yarn? Would love to know.

They can, but it's super easy to have outdated lock file in such case. Personally I just usually disable locks completely in library setting because they seem (to me) to bring more chaos than good.

PR #14, opened 2018-08-29
no need to use else statement

sunnibfgi:
Infer by conditional judgment, else statement does not seem to be necessary here.


ryanflorence:
thanks for taking the time but I like the else statements so my editor can help me jump between if/else, also `refs` is intended to be undefined, not an object if they aren't specified :)

PR #16, opened 2018-08-29
Specify missing values for aria-hidden attribute

yuheiy:
`aria-hidden` is enumerated attribute.

See: https://www.w3.org/TR/wai-aria-1.1/#aria-hidden

### DIFF
--- a/www/src/pages/menu-button.mdx
+++ b/www/src/pages/menu-button.mdx
@@ -12,10 +12,10 @@ import Helmet from "react-helmet"
 
 # MenuButton (Dropdown)
 
-[Menu](#menu) - 
-[MenuButton](#menubutton) - 
-[MenuList](#menulist) - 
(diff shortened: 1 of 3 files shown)


ryanflorence:
In React if you do `<span aria-hidden/>` it will render `<span aria-hidden="true"/>` to the DOM.

But it's probably good to be more explicit since for some people this is the first time they've really even seen the attribute.

I'll merge this if you update it w/ the comments I'm about to make :)

www/src/pages/menu-button.mdx
```diff
@@ -12,10 +12,10 @@ import Helmet from "react-helmet"
 
 # MenuButton (Dropdown)
 
-[Menu](#menu) -
```
ryanflorence: I *think* this trailing whitespace needs to be there so that there's a space between the characters.

PR #17, opened 2018-08-29
Fix memory leak with event handler in WindowSize component.

pvormittag:
Simple fix for unintentionally calling `addEventListener` inside `removeEvent`.


ryanflorence:
woah, thanks!

PR #38, opened 2018-09-04
Handle menu collision at bottom, and at bottom and right

brendancarney:
Adds handling for the menu colliding at the bottom, and colliding the right and the bottom. 

Let me know if you have any feedback or if I should add any examples or tests. 

Thanks for taking the time to review! 

**Bottom**
screen shot 2018-09-03 at 7 41 44 pm

**Right (already included)**
screen shot 2018-09-03 at 7 41 17 pm

**Right and Bottom**
screen shot 2018-09-03 at 7 41 29 pm

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -393,31 +393,25 @@ let getStyles = (buttonRect, menuRect) => {
     styles.minWidth = buttonRect.width;
   }
 
-  let collisionRight = window.innerWidth < buttonRect.left + menuRect.width;
-  let collisionBottom = window.innerHeight < buttonRect.top + menuRect.height;
-  let collisionRightAndBottom = collisionRight && collisionBottom;
+  let collisions = {
+    top: buttonRect.top - menuRect.height < 0,
+    right: window.innerWidth < buttonRect.left + menuRect.width,
+    bottom: window.innerHeight < buttonRect.top + menuRect.height,
+    left: buttonRect.left - menuRect.width < 0
+  };
 
-  if (collisionRightAndBottom) {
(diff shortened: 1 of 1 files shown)


brendancarney:
What happens: The menu direction goes up
What should happen: Not sure. I *think* it should probably default to normal behavior in that case (menu goes down)

I'm thinking that a right & left collision should be consistent with that also.

ryanflorence:
I agree, should go down, the window scrollHeight will expand and focus moves to the items so it will remain usable.

If you can get that working I'll merge this stuff.  It looks great!

Side note: Another thing to worry about is right-to-left languages, where the default should be reversed horizontally, but I haven't decided on the general way to branch for RTL, so don't worry about it yet, just start thinking about it 😋I'm thinking we'll have a RTLContext to read from but for now I'm trying to avoid any top level `<Provider>`'s

brendancarney:
Updated. Let me know if you have different ideas on the implementation. There were so many different ways to do this. I _think_ this should be relatively easy to update for RTL when we get there.

rafaelchiti:
Out of curiosity, any specific reason to not use something like popper js? Curious about the decision of manually handling this logic (which is complicated, positioning in general and collision). Thanks!

seancurtis:
I'd echo @rafaelchiti's comments regarding using something like popper.js – there are so many edge cases with this kind of stuff, and something like Popper has them solved already.

What happens if your menu is in a scroll parent? Gets clipped right? This is where you need position: fixed and a bunch of scroll listeners. I'd highly recommend taking a look at using react-popper if you're positioning things.

Issue #35, opened 2018-09-01
Marking required fields in documentation


Glazy:
For the component documentation it might be nice if there was a way to mark props as required. This could be added as a column to the `Props` table for each component maybe?

ryanflorence:
Absolutely, and while we're at it actually add `propTypes` to all of the components in the code 🙈

PR #44, opened 2018-09-04
Prop types menu button

Glazy:
Add `propTypes` for `menu-button` and update the associated documentation with a new `required` column for each prop.

Gets some way towards closing: #35 🎉

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -4,7 +4,7 @@ import { Link } from "@reach/router";
 import Rect from "@reach/rect";
 import WindowSize from "@reach/window-size";
 import Component from "@reach/component-component";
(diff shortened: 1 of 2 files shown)


ryanisinallofus:
Is the idea to just add a column here for #35?

screen shot 2018-09-04 at 5 08 50 pm

Glazy:
Yeah @ryanisinallofus :+1:

packages/menu-button/src/index.js
```diff
@@ -228,6 +229,15 @@ let MenuListImpl = React.forwardRef(
   )
 );
 
+MenuListImpl.propTypes = {
```
ryanflorence: do we need these since this is just an implementation detail?

ryanflorence:
lovely! thank you

PR #45, opened 2018-09-04
Opening external links in new tab

olezhek85:
Added target attribute to open Spectrum Community and GitHub links in new tab.

### DIFF
--- a/www/src/components/layout.js
+++ b/www/src/components/layout.js
@@ -29,6 +29,7 @@ let NavLink = React.forwardRef(
         ref={ref}
         className="NavLink"
         target="_blank"
+        rel="noopener"
         {...props}
       />
     ) : (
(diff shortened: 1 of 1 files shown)


olezhek85:
Added.

ryanflorence:
Any reason people can't just hold command or control and click the link? What's the justification for opening in a new window?

MaxymSemenykhin:
> Any reason people can't just hold command or control and click the link?
 Why I as user should press control if it's obviously should be an external link in new tab?

For example https://reactjs.org/ bootom links 

image


They all contain `target="_blank" ` , so for me as regular user it is obviously that those links should have`target="_blank"` any valid arguments that they shouldn't?

olezhek85:
It's a common practice to open external links in a new tab, isn't it?

ryanflorence:
I've never known how to decide these things, but there are two confetti emojis on the last comment so that's enough for me 😂

Issue #20, opened 2018-08-30
dialog: handle label automatically


ryanflorence:
Dialogs should have `aria-label` or `aria-labelled-by`, thinking we could help people out by requiring you render a `<DialogLabel>This is the label</DialogLabel>`.

If a dialog renders and doesn't find a `DialogLabel` inside, it could warn in the console.

Internally it could handle generating the ids for the developer and hook up the `aria-labelled-by`.

giuseppeg:
I like the `accessibilityLabel` prop approach of React Native[ for Web]

everdimension:
`<DialogLabel />` component kinda implies that you can render arbitrary markup inside of it, which, as I understand, is not the idea.

chancestrickland:
`accessibilityLabel` Feels like a nice approach here to adopt as a standard across the project. Nice bonus that it will be familiar for native folks.

kylemh:
Was curious if leveraging AlertDialogLabel is valid within a regular Dialog component?

PR #52, opened 2018-09-07
[WIP] Require presence of `DialogLabel` in `Dialog`

nonnontrivial:
I've tried my hand at what is recommended in https://github.com/reach/reach-ui/issues/20.

I'm wondering if this approach – which I've more or less mimed from `AlertDialog` – is what we're looking for. And whether the implementation in `AlertDialog` can now be removed (because it imports the standard dialog) and the examples updated. (This breaks any example featuring a `Dialog` without a `DialogLabel`.)

### DIFF
--- a/packages/dialog/src/index.js
+++ b/packages/dialog/src/index.js
@@ -2,7 +2,9 @@ import React from "react";
 import Component from "@reach/component-component";
 import Portal from "@reach/portal";
 import { checkStyles, wrapEvent } from "@reach/utils";
+import { Consumer as IdConsumer } from "@reach/utils/lib/IdContext";
 import createFocusTrap from "focus-trap";
+import invariant from "invariant";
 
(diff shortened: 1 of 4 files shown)


packages/dialog/src/index.js
```diff
@@ -71,36 +73,54 @@ let DialogOverlay = React.forwardRef(
...
+                refs={{
+                  labelId: `dialog-${genId()}`,
+                  overlayNode: null,
+                  contentNode: null
+                }}
+                didMount={({ refs }) => {
+                  portalDidMount(refs, initialFocusRef);
+                  invariant(
+                    document.getElementById(refs.labelId),
+                    `@reach/dialog: You must render a \`<DialogLabel>\`
```
Glazy: Maybe this should be? 

> `<Dialog />` must render a `<DialogLabel />`

Issue #54, opened 2018-09-11
MenuItem doesn't take react-router's Link component 


nicolasschabram:
Hey there,

I was trying out `MenuButton` and noticed that `MenuLink` doesn't take react-router's `Link` as `component` prop:

> Warning: Failed prop type: Invalid prop `component` supplied to `ForwardRef`.

Is that because of what's written [in the docs](https://ui.reach.tech/menu-button#menulink-component)?

> If other routers' Link component uses the React.forwardRef API, you can pass them in as well. If they don’t it won't work because we will not be able to manage focus on the element the component renders.

If so, am I missing an obvious work-around (we're not quite ready yet to go all in with `reach-router`)? Otherwise, would it make sense for the docs to state the component's incompatibility with the widest used router a bit more explicitly?

Thanks a lot for building this!
Nicolas

----

Broken example:

```jsx
import { Menu, MenuButton, MenuLink, MenuList } from "@reach/menu-button";
import { Link } from "react-router-dom";

const Nav = () => (
  <Menu>
    <MenuButton>My Button</MenuButton>
    <MenuList>
      <MenuLink component={Link} to="/route">Click me!</MenuLink>
    </MenuList>
  </Menu>
);
```

adamduncan:
Wouldn't using `MenuLink`'s `onClick` prop to manually push to React Router's history work? E.g.

```jsx
<MenuLink
  to="/my/route"
  onClick={() => this.props.history.push('/my/route')}
>Link to url</MenuLink>
```

That way it's still an accessible anchor tag with an href. We intercept its click event, instead of running into trouble trying to pass React Router's `<Link>` as `component`?

I can imagine a lot of users are in @nicolasschabram's (and my) situation, and would benefit from a workaround solution while we consider migrating our projects to `reach-router`.

Cheers

pietervanneste:
hi @ryanflorence, any update on this?

lucky-kim:
@ryanflorence Any update?

schabluk:
Possible workaround (works with new Materia-UI v 4).

```javascript
const NavigationLink = React.forwardRef(({ nested = false, ...props }, ref) => {
  return <Link {...props} ref={ref} />
})
```

tofful:
> Possible workaround (works with new Materia-UI v 4).
> 
> ```js
> const NavigationLink = React.forwardRef((props, ref) => {
>   return <Link {...props} ref={ref} />
> })
> ```

it doesn't work with typescript.

[...]

PR #55, opened 2018-09-11
Update MenuButtonLink compatibility info

Glazy:
Someone raised in an issue (#54) that it might just be easier if we clearly
state that `react-router`'s `<Link />` component doesn't work as of the
latest version.


ryanflorence:
Thanks, but It will be soon :)

PR #57, opened 2018-09-11
Add PropTypes for AlertDialog, Dialog, and WindowSize; fix warnings in Rect example

filipdanic:
A small help towards the ongoing work in #35.

Also fixes warning that show up in the Rect examples.

### DIFF
--- a/packages/rect/examples/basic.example.js
+++ b/packages/rect/examples/basic.example.js
@@ -8,7 +8,7 @@ export let Example = () => (
     {({ ref, rect }) => (
       <div>
         <pre>{JSON.stringify(rect, null, 2)}</pre>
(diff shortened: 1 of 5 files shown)


packages/rect/examples/basic.example.js
```diff
@@ -8,7 +8,7 @@ export let Example = () => (
     {({ ref, rect }) => (
       <div>
         <pre>{JSON.stringify(rect, null, 2)}</pre>
-        <textarea>resize this</textarea>
```
filipdanic: fixes: `Warning: Use the `defaultValue` or `value` props instead of setting children on <textarea>.`

ryanflorence:
thanks!

PR #59, opened 2018-09-13
Begin dialog tests

nonnontrivial:
I've begun writing some tests for `Dialog`. Accommodating the portal may require workarounds (see https://github.com/facebook/react/issues/11565).


ryanflorence:
thanks!

PR #60, opened 2018-09-13
Add prop-types peer dependency to packages

spentacular:
Hey Ryan,

Thanks for taking on the challenge and putting so much thought into these accessible components. I'm super pumped about the project and look forward to see how it grows.

I was checking out on Bundlephobia the size of `@reach/component-component` and it wasn't able to generate the stats because `prop-types` was missing:

screen shot 2018-09-13 at 9 27 52 am


ryanflorence:
thanks!

Issue #37, opened 2018-09-03
Type definitions


Glazy:
It would be great to get some TypeScript type definitions written for each component.

ryanflorence:
I'm all for it, but we're going to need somebody dedicated to that, because I don't know typescript well enough to do it myself, or even review the work of others for it, and I'm not interested enough in typescript to learn it 😬(I know, I should be, and I would love it)

Miklet:
Just to have better look at current state of type definitions for all packages:

- [ ] alert-dialog
- [x] alert
- [ ] component-component
- [ ] dialog
- [x] menu-button
- [ ] portal
- [ ] rect
- [x] skip-nav
- [ ] utils
- [x] visually-hidden
- [x] windows-size

hedgerh:
You start on `component-component` yet?  I've never done a 3rd party typedef before but im about to use `Component`, so I figured I may as well start writing one out

hedgerh:
Could use some feedback on my `@reach/component-component` typedefs.  I believe it's complete, but it may not be as elegant as it could be. https://github.com/reach/reach-ui/pull/105

hedgerh:
Sorry to keep pinging this issue, but I went ahead and started all of the other remaining types.  They are part of the same PR I linked above.

SleeplessByte:
```typescript
declare module '@reach/tabs' {
  import { ComponentType } from "react"

  export interface BaseTabProps<A> {
    children?: React.ReactNode
    as?: ComponentType
    rest?: A
  }

  export interface TabsProps<A = React.HTMLAttributes<HTMLDivElement>> extends BaseTabProps<A> {
    children: React.ReactNode

    defaultIndex?: number
    index?: number
    onChange?(index: number): void
  }
  export class Tabs extends React.Component<TabsProps> {}

  export type TabListProps<A = React.HTMLAttributes<HTMLDivElement>> = BaseTabProps<A>
  export class TabList extends React.Component<TabListProps> {}

  export type TabPanelsProps<A = React.HTMLAttributes<HTMLDivElement>> = BaseTabProps<A>
  export class TabPanels extends React.Component<TabPanelsProps> {}

  export type TabPanelProps<A = React.HTMLAttributes<HTMLDivElement>> = BaseTabProps<A>
  export class TabPanel extends React.Component<TabPanelProps> {}

  export interface TabProps<A = React.HTMLAttributes<HTMLDivElement>> extends BaseTabProps<A> {
    disabled?: boolean
  }
[...]

PR #63, opened 2018-09-14
Add TS definitions for visually-hidden package

Miklet:
Hello

It's me again 😄 

After #62 (which I hope it is going to be merged) I started to continue work started by @Glazy at better TS definitions for packages in this repo #37

This PR adds TS definitions for `visually-hidden` package. I put them in it's own folder `typings` but not sure about it since @Glazy was putting there in root folder 😅 Let me know if I should change it 😉 

Cheers


packages/visually-hidden/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Render text that is announced to screen readers but visually hidden.",
   "main": "index.js",
   "module": "es/index.js",
+  "typings": "typings/index.d.ts",
```
Glazy: I think to remain consistent with my PRs (unless anyone feels strongly otherwise) let's move this file into the root of the package :+1:

Miklet: Actually I did that first but than I thought that another folder would be a good idea 😅 But I agree it is better to keep consistence 👌

Glazy: Cool, well as soon as you make that change I will approve this for you :+1: 🙂

hasparus:
@Miklet @Glazy 
I think types for VisuallyHidden are not published
image

PR #68, opened 2018-09-18
Implement `<Disclosure />` component

Glazy:
Hey everyone :wave:

Had some free time so I thought I would get to work on implementing a new component for Reach. I decided to go with the [Disclosure component](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure).

If you spot any accessibility concerns then let me know but it shouldn't have any as it matches the spec defined on W3.org!

I also may have strayed slightly away from convention seeing as this is my first solo Reach component so if I have let me know and I will fix it :+1:


packages/disclosure/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+        {state.isExpanded && (
+          <div data-reach-disclosure-content>{children}</div>
+        )}
+      </React.Fragment>
+    )}
+  </Component>
+);
+
+Disclosure.propTypes = {
+  buttonLabel: PropTypes.string.isRequired
```
VinceMalone: I'm not aware of any [component (API) design guidelines] for this library, but the trend does seem to teeter more on the pattern of compound components. With the [button content] being a prop, that's a string, the consumer of this component would be limited with how they want to render the button; e.g. the following wouldn't be possible:

```jsx
<Disclosure
  buttonLabel={
    <React.Fragment>
      More <strong className="flashy">info</strong>
    </React.Fragment>
  }
>
  <p>content</p>
</Disclosure>
```

At the very least, I'd consider making the prop-type `node`; but perhaps using the context API and compound components pattern, you could mimic the API of the native HTML [`details`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element:

```jsx
import { Disclosure, DisclosureSummary, DisclosureContent } from '@reach/Disclosure';

// ...

<Disclosure>
  <DisclosureSummary>
    More <strong className="flashy">info</strong>
  </DisclosureSummary>
  <DisclosureContent>content</DisclosureContent>
</Disclosure>
```

Glazy: Hey @VinceMalone,

I think you've made a good point and I would be more than happy to update the API. I think a render prop could work quite well in this use case for the `buttonContent`. Would be interesting to know what @ryanflorence thinks :+1:

chancestrickland:
Hey @Glazy! Jumping in here, as I'm new on the Reach team and trying to clean out the backlog. We are actively working on an `Accordion` component that I believe will cover much of the disclosure's behavior per the ARIA spec. That will be rolling out for beta testing and contributions over the next couple of weeks. We'd love your feedback and contributions.

My apologies that this went stale before we could get feedback on the API to you. I'm going to close the PR for now and I'll make sure to tag you once the accordion branch is up and ready for testing.

PR #69, opened 2018-09-19
Add onDismiss to AlertDialog examples

danieltott:
I completely missed this highly useful prop my first time around. I was thinking adding it to the basic examples help others not miss it as well!

### DIFF
--- a/packages/alert-dialog/examples/basic.example.js
+++ b/packages/alert-dialog/examples/basic.example.js
@@ -21,7 +21,10 @@ export let Example = () => (
         </button>
 
         {state.showDialog && (
-          <AlertDialog leastDestructiveRef={refs.close}>
+          <AlertDialog
+            leastDestructiveRef={refs.close}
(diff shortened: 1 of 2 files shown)


danieltott:
@Glazy You might be right about that. I was a little unsure which way to go. 

It seems like with `AlertDialog`, that `onDismiss` and `leastDestructiveRef` go hand in hand, since the docs say that they should call the same event handler, and `leastDestructiveRef` is required. But, I'm not entirely sure about the accessibility requirements or guidelines related to closing dialogs like this. If you think I should break it out into its own story, I'd be happy to do that.

danieltott:
@ryanflorence I think that removing `onDismiss` and having escape and clicking outside the modal just to the same as the `leastDestructiveRef` click handler would be perfectly fine. This note in the docs:

> IMPORTANT: Ensure that onDismiss and the click handler of the leastDestructiveRef are identical!

Seems to say the same thing.

PR #74, opened 2018-10-04
fix react-dom imports

kentcdodds:
react-dom does not use named exports and when rollup attempts to bundle
these files it gives the following error:

```
[!] Error: 'render' is not exported by ../../node_modules/react-dom/index.js
https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
node_modules/@reach/alert/es/index.js (19:9)
17:
18: import React from "react";
19: import { render } from "react-dom";
             ^
20: import Component from "@reach/component-component";
21: import VisuallyHidden from "@reach/visually-hidden";
```

### DIFF
--- a/packages/alert-dialog/src/index.js
+++ b/packages/alert-dialog/src/index.js
@@ -1,10 +1,10 @@
-import React, { createContext, forwardRef } from "react";
+import React from "react";
 import Component from "@reach/component-component";
 import { DialogOverlay, DialogContent } from "@reach/dialog";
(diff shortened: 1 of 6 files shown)


ryanflorence:
Huh... well alright.

TrySound:
@kentcdodds You could just specify namedExports in commonjs plugin.

kentcdodds:
Yeah, should've done that. But my preferred solution would be: https://github.com/facebook/react/issues/13272

TrySound:
Work in progress in https://github.com/facebook/react/pull/13321

PR #78, opened 2018-10-05
Add Render Callback to Menu

### LINKED ISSUE #76: MenuList Render Callback
In #28 you mention adding a render callback for exposing `isOpen` as well as enabling exit animations.    I could take a look at implementing this change.

Any known gotchas I should consider?

nathanforce:
Resolves #76

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -78,12 +78,18 @@ let Menu = ({ children }) => (
     didUpdate={manageFocusOnUpdate}
     getSnapshotBeforeUpdate={checkIfAppManagedFocus}
   >
-    {context => <Provider value={context}>{children}</Provider>}
+    {context => (
+      <Provider value={context}>
+        {typeof children === "function"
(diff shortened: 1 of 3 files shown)


packages/menu-button/src/index.js
```diff
@@ -78,12 +78,24 @@ let Menu = ({ children }) => (
     didUpdate={manageFocusOnUpdate}
     getSnapshotBeforeUpdate={checkIfAppManagedFocus}
   >
-    {context => <Provider value={context}>{children}</Provider>}
+    {context => (
+      <Provider value={context}>
+        <Consumer>
```
ryanflorence: Should be able to skip the consumer and use `context.state.isOpen` yeah?

```jsx
<Provider value={context>
  {typeof children === 'function' ? children({ isOpen: context.state.isOpen }) : children }
</Provider>
```

nathanforce: Done.

PR #79, opened 2018-10-05
Add Support for Non menu-button Children

### LINKED ISSUE #75: <MenuList> children other than <Menu* />
I saw in the Notes.md for menu-button you have
```
- Ignore non-menu-items for custom rendering
```

I'd be happy to take a stab at this, as its a feature I need anyhow.  Wanted to pick your brain around what you had in mind.  Should we check `child.type`? First hunch is to leave any non `<Menu* />` children untouched

nathanforce:
Closes #75

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -141,106 +141,6 @@ MenuButton.propTypes = {
   children: node
 };
 
-////////////////////////////////////////////////////////////////////////
-
-let MenuList = React.forwardRef((props, ref) => (
-  <Consumer>
-    {({ refs, state, setState }) =>
-      state.isOpen && (
-        <Portal>
-          <WindowSize>
-            {() => (
-              <Rect>
-                {({ rect: menuRect, ref: menuRef }) => (
-                  <div
-                    data-reach-menu
-                    ref={menuRef}
-                    style={getStyles(state.buttonRect, menuRect)}
-                  >
-                    <MenuListImpl
-                      {...props}
-                      setState={setState}
-                      state={state}
-                      refs={refs}
-                      ref={ref}
-                    />
-                  </div>
-                )}
-              </Rect>
-            )}
-          </WindowSize>
-        </Portal>
-      )
-    }
-  </Consumer>
-));
-
-MenuList.propTypes = {
-  children: node
-};
-
-let MenuListImpl = React.forwardRef(
-  ({ refs, state, setState, children, onKeyDown, onBlur, ...rest }, ref) => (
-    <div
-      data-reach-menu-list
-      {...rest}
-      role="menu"
-      aria-labelledby={state.buttonId}
-      tabIndex="-1"
-      ref={node => {
-        refs.menu = node;
-        ref && ref(node);
-      }}
-      onBlur={event => {
-        if (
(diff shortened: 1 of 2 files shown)


nathanforce:
@ryanflorence Would you consider replacing the `cloneElement` approach with a context based registration approach to support arbitrarily deep `MenuItem` nesting? Or is that in violation of ARIA? Do `menuitems` need to be direct children of `menu`?

packages/menu-button/src/index.js
```diff
@@ -369,6 +271,120 @@ MenuLink.propTypes = {
...
+            setState(close);
+          }
+        }}
+        onKeyDown={wrapEvent(onKeyDown, event => {
+          if (event.key === "Escape") {
+            setState(close);
+          } else if (event.key === "ArrowDown") {
+            event.preventDefault(); // prevent window scroll
+            let nextIndex = state.selectionIndex + 1;
+            if (nextIndex !== focusableChildren.length) {
```
nathanforce: Notable change `React.Children.count(children) -> focusableChildren.length` is lost in the noise here.

I had to move MenuList below MenuItem/MenuLink in order to define and subsequently use `focusableChildrenTypes`

packages/menu-button/src/index.js
```diff
@@ -369,6 +271,120 @@ MenuLink.propTypes = {
...
+        })}
+      >
+        {React.Children.map(children, (child, index) => {
+          if (isFocusableChildType(child)) {
+            let focusIndex = focusableChildren.indexOf(child);
+
+            return React.cloneElement(child, {
+              setState,
+              state,
+              index: focusIndex,
```
nathanforce: Here we pass `focusIndex` instead of actual index

packages/menu-button/src/index.js
ryanflorence: Cool, this is exactly what I had in mind. Going to pull this code down and mess around w/ it a bit.

ryanflorence:
> Would you consider replacing the cloneElement approach with a context based registration approach

Yeah, I try to avoid `cloneElement` but I avoid context registration even more because you can't server render that stuff (assuming we're talking about the same strategy).

But since these menus can't be rendered open initially anyway I say we do it. You wanna take a stab at that?

ryanflorence:
(Also, think you can resolve these conflicts and then we can get this stuff merged? It's actually not very usable w/o this PR!)

nathanforce:
> But since these menus can't be rendered open initially anyway I say we do it. You wanna take a stab at that?

Sure thing. I'll resolve these conflicts and take a stab at the context changes in another PR

nathanforce:
This should be good to go 👍

ryanflorence:
Thanks! I added a bit more to clear the weird highlight when you mouse out of a menu item and into a non-menu item in the next commit badc98c79167abc80357faebcec543d0d83592bb

nathanforce:
@ryanflorence I don't think that is quite right. I'm getting a runtime error when trying out master locally.

To reproduce: hover a non-menu child and then ArrowUp.

You've only added the `MouseLeave` logic for `MenuItem`, so any other children (including `MenuLink`, I believe) will not reset the index as intended.

PR #81, opened 2018-10-09
Make tabs, tabbar, tab, and tabpanel components to handle tab navigat…

cdevine49:
…ion (arrow keys), roles, and arias (controls, selected, labelledby)

The basic structure of the components is based on (almost exactly the same as) something @ryanflorence put online a while back.

There was some debate about whether the aria roles are actually good and if you should hit tab to move between tabs instead of / in addition to using arrow keys.  I opted to include roles and only allow arrow key navigation because the project i looked at that went the other way on those was tough to navigate.

This is pretty bare right now ie. no tests, styles, documentation, but if this looks like a good start I can work on all that.

### DIFF
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -151,7 +151,7 @@ let TabPanels = ({ activeIndex, tabIds, ...props }) => {
   );
 };
 
-TabPanels.prop = {
+TabPanels.propTypes = {
   activeIndex: number.isRequired,
   tabIds: arrayOf(string).isRequired
 };
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -1,5 +1,5 @@
 import React, { Component, Fragment } from "react";
-import { string } from "prop-types";
+import { arrayOf, bool, func, number, string } from "prop-types";
 
 class Tabs extends Component {
   constructor(props) {
@@ -43,12 +43,16 @@ class Tabs extends Component {
   }
 }
 
+Tabs.propTypes = {
+  activeIndex: number
+};
+
 Tabs.defaultProps = {
   activeIndex: 0
 };
 
 let TabBar = ({ activeIndex, onClick, panelIds, ...props }) => {
-  const maxIndex = props.children && props.children.length - 1;
+  const maxIndex = React.Children.count(props.children);
 
   return (
     <ul role="tablist" {...props}>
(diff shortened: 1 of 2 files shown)


packages/tabs/src/index.js
```diff
@@ -0,0 +1,140 @@
+import React, { Component, Fragment } from "react";
+import { string } from "prop-types";
+
+class Tabs extends Component {
```
dennisja: Tabs is missing propType validation as well

packages/tabs/src/index.js
```diff
@@ -0,0 +1,140 @@
...
+      </li>
+    );
+  }
+}
+
+Tab.propTypes = {
+  id: string.isRequired
+};
+
+let TabPanels = ({ activeIndex, tabIds, ...props }) => {
```
dennisja: Looks like TabPanels is missing propType validation as well

packages/tabs/src/index.js
```diff
@@ -0,0 +1,140 @@
...
+        >
+          {props.children}
+        </a>
+      </li>
+    );
+  }
+}
+
+Tab.propTypes = {
+  id: string.isRequired
```
dennisja: What about other props like onClick, index, onKeyDown, onNextTab...?

packages/tabs/src/index.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+  }
+}
+
+Tabs.defaultProps = {
+  activeIndex: 0
+};
+
+let TabBar = ({ activeIndex, onClick, panelIds, ...props }) => {
+  const maxIndex = props.children && props.children.length - 1;
```
dennisja: This can be undefined, won't it be better to use `React.Children.count(children)`?

packages/tabs/src/index.js
```diff
@@ -137,4 +151,9 @@ let TabPanels = ({ activeIndex, tabIds, ...props }) => {
   );
 };
 
+TabPanels.prop = {
```
dennisja: Did you mean TabPanels.propTypes?

ryanflorence:
Hey, thanks for taking the time to add this, I really appreciate it.

I've actually got some tabs coming that are built with hooks and support some unique Suspense use-cases. I'll make sure you add you to review in the PR!

PR #85, opened 2018-10-16
Update VisuallyHidden to render span

crosscompile:
First off, thanks for all the work you're doing with Reach!

I was reading the [spec for `button`](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) and noticed that it specifies it's contents must be phrasing content which does not include `div`.

I'm not sure if there are any side effects from rendering a `div` within a `button`, but I figured it would be best to follow the spec for `VisuallyHidden` as it can be used within elements accept phrasing content.


ryanflorence:
Woah ... I thought this was a span! Thanks!

ryanflorence:
cherry-picked as cf34a79d2e95a1dce05ffde3359357eb75f99bc0

Issue #87, opened 2018-10-24
[Dialog] Add prop to specify the root of the dialog


everdimension:
I noticed that currently you cannot specify the root for the dialog to be rendered into https://github.com/reach/reach-ui/blob/master/packages/portal/src/index.js#L10

The `body` element is hardcoded to the <Portal /> component.

I would very much like to see this feature. It's very important for apps that get integrated into other apps.

Would you accept a PR implementing this?

everdimension:
@denkristoffer No, I haven't yet. I wanted to make sure that the project wants this first. But I guess it wouldn't hurt if you want to send a PR before that

denkristoffer:
I currently need this functionality so I would have to build something to do it either way 🙂 I've started work on this in #88, would like to hear your thoughts if you have any.

ryanflorence:
let's continue the conversation in #88

PR #88, opened 2018-11-05
Portal: Add container prop to allow rendering portal in a specific DOM node

denkristoffer:
This PR is an initial take on how to address https://github.com/reach/reach-ui/issues/87

Sometimes it'd be helpful to be able to render `<Dialog />` in other DOM nodes than `document.body`. This PR enables this by allowing the user to pass a `container` prop with a React ref pointing to the DOM node that they want to render the portal containing the `<Dialog />` in.

Edit: I forgot to handle `aria-hidden`.

### DIFF
--- a/packages/portal/src/index.js
+++ b/packages/portal/src/index.js
@@ -2,16 +2,28 @@ import React from "react";
 import ReactDOM from "react-dom";
 import Component from "@reach/component-component";
 
-let Portal = ({ children, type = "reach-portal" }) => (
+let Portal = ({
+  children,
+  container = document.body,
+  type = "reach-portal"
+}) => (
   <Component
     getRefs={() => ({ node: null })}
     didMount={({ refs, forceUpdate }) => {
+      let containerNode = container.hasOwnProperty("current")
+        ? container.current
+        : container;
       refs.node = document.createElement(type);
-      document.body.appendChild(refs.node);
+      containerNode.appendChild(refs.node);
       forceUpdate();
     }}
     willUnmount={({ refs: { node } }) => {
-      document.body.removeChild(node);
+      let containerNode = container.hasOwnProperty("current")
+        ? container.current
+        : container;
+      if (containerNode) {
+        containerNode.removeChild(node);
+      }
     }}
     render={({ refs: { node } }) => {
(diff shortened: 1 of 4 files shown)


packages/portal/src/index.js
```diff
@@ -2,16 +2,28 @@ import React from "react";
 import ReactDOM from "react-dom";
 import Component from "@reach/component-component";
 
-let Portal = ({ children, type = "reach-portal" }) => (
+let Portal = ({
+  children,
+  container = document.body,
```
everdimension: This might break server-side rendering.

ryanflorence:
My hesitation here is that in order to have dialogs on top of dialogs (even though it's a design sin, we still do it all the time) we need to ensure that the document order is correct, that the newest dialogs render later in the document than the ones before.

Can you expand further on this?

>  It's very important for apps that get integrated into other apps.

everdimension:
> Can you expand further on this?

Yeah, basically I meant something like this:

screen shot 2019-01-04 at 15 26 12

It makes sense when your app is something like a charting solution or a visual builder of some sort (e.g. some UI that serves one isolated purpose, but also influences the external app). You could argue that putting such app in an `iframe` would solve the problem, but using an `iframe` is not always desirable.

You could also imagine some smaller widget for which it wouldn't make sense to block the whole page.

ryanflorence:
Why would the modal not cover the "external website" nav? My understanding of a modal dialog is to interrupt the previous UI, not display them together.

andreiglingeanu:
Is there anything we can do to make this prop happen? I see that the PR is a bit outdated but I'm more than happy to work towards getting it ready to be merged.

Now to describe a little bit the use case for this: I have fixed width sidebar in a dashboard and a main working area. What I'm trying to achieve is centering the overlays relative to the working area -- that'll look a bit cleaner from the design standpoint. Exactly what @everdimension described.

To address the dialogs stacking one on top of another: I'll make sure to always pass the same `container` prop for all overlays inside the app. That way they can happily stack, but cover only the working area of the app. Even more, dialogs stacking is very unintuitive for users, and everything that goes >1 level deep is bad, in my opinion.

PR #91, opened 2018-11-14
Allow users to override position using setPosition on the MenuList

krzystof:
Allows users to pass a `setPosition` props to the `MenuList` components to overrides styles of the menu.

That will allow users to position the menu list wherever they want (centered, to the right of the dropdown, above, ...)


ryanflorence:
It currently has some collision detection w/ the window edges, this would ruin that.

Perhaps you can do it with CSS?

If anything, maybe just use a big hammer and provide a `getPosition` prop that takes a function that passes in whatever information might be important?

PR #92, opened 2018-11-14
Add TS types for @reach/menu-button

krzystof:
Add types for the menu-button: #37


Pau1fitz:
@ryanflorence any idea when these will be merged?

pluma:
Can we have an updated release? Looks like the latest version still doesn't ship these.

PR #93, opened 2018-11-14
Add missing forwardRef for MenuList

krzystof:
`MenuListImpl` contains `ref && ref(node)`, so I think the intention was to allow users to pass in a ref to the wrapping div.

The `React.forwardRef` is missing on the `MenuList` component.


ryanflorence:
Thanks ... I *think* that's the element people would expect?

Issue #96, opened 2018-12-09
menu-button: add `component` prop (on MenuButton)


VinceMalone:
Like `MenuLink`, it would be nice to add a `component` prop on `MenuButton`. This would allow [abstractions through components] to be used with `MenuButton`; for example:

1. Styles (that don't, **or can't**, rely on the `[data-*]` approach)

```jsx
<MenuButton component={MyFancyButton} theme="really-fancy" />
```

2. Button components that care about various events (like `focus`) as they relate to the MenuButton components; e.g. show a tooltip on focus.

```jsx
<MenuButton component={ButtonThatDoesShitOnFocus} />
```

Pau1fitz:
I think I have a similar use case, or else perhaps it is possible to do this already.

I have a logout component that looks like the following:

```js
import React from 'react';
// this has come from from useReducer
import { useDispatch } from '../store';

const Logout = () => {
	const dispatch = useDispatch();
	const performLogout = (): void => {
		dispatch({
			type: 'LOGOUT_SUCCESS',
		});
	};

	return (
		<div onClick={performLogout}>Logout</div>
	);
};

export default Logout;
```

I then import this into my `Menu`.

```js
import React from 'react';
import {Menu, MenuList, MenuButton, MenuItem } from "@reach/menu-button";

const Dropdown = () => {
	return(
		<Menu>
			<MenuButton>
				Select
			</MenuButton>
			<MenuList>
				<MenuItem onSelect={() => alert("Settings")}>Settings
				</MenuItem>
				<MenuItem onSelect={() => performLogout()}>
					Logout
				</MenuItem>
			</MenuList>
		</Menu>
	);
};

export default Dropdown;
```

Is there anyway to call the method defined in my `Logout` component?

ryanflorence:
closing in favor of the PR, thanks!

PR #97, opened 2018-12-09
menu-button: add `component` prop to MenuButton

VinceMalone:
Addresses https://github.com/reach/reach-ui/issues/96

> Like `MenuLink`, it would be nice to add a `component` prop on `MenuButton`. This would allow [abstractions through components] to be used with `MenuButton`.

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -99,57 +99,68 @@ Menu.propTypes = {
 };
 
 ////////////////////////////////////////////////////////////////////////
-let MenuButton = React.forwardRef(({ onClick, onKeyDown, ...props }, ref) => (
-  <Consumer>
-    {({ refs, state, setState }) => (
-      <Rect
-        observe={state.isOpen}
-        onChange={buttonRect => setState({ buttonRect })}
-      >
-        {({ ref: rectRef }) => (
-          <button
-            id={state.buttonId}
-            aria-haspopup="menu"
-            aria-expanded={state.isOpen}
-            data-reach-menu-button
-            type="button"
-            ref={node => {
-              rectRef(node);
-              ref && ref(node);
-              refs.button = node;
-            }}
-            onMouseDown={event => {
-              if (state.isOpen) {
-                setState({ closingWithClick: true });
-              }
-            }}
-            onClick={wrapEvent(onClick, event => {
-              if (state.isOpen) {
-                setState(close);
-              } else {
-                setState(openAtFirstItem);
-              }
-            })}
-            onKeyDown={wrapEvent(onKeyDown, event => {
-              if (event.key === "ArrowDown") {
-                event.preventDefault(); // prevent scroll
-                setState(openAtFirstItem);
-              } else if (event.key === "ArrowUp") {
-                event.preventDefault(); // prevent scroll
-                setState(openAtFirstItem);
-              }
-            })}
-            {...props}
-          />
-        )}
-      </Rect>
-    )}
-  </Consumer>
-));
+let MenuButton = React.forwardRef(
+  (
+    {
+      onClick,
+      onKeyDown,
+      asAccessibleButtonComponent: Comp = "button",
+      ...props
+    },
+    ref
+  ) => (
+    <Consumer>
+      {({ refs, state, setState }) => (
+        <Rect
+          observe={state.isOpen}
+          onChange={buttonRect => setState({ buttonRect })}
+        >
+          {({ ref: rectRef }) => (
+            <Comp
+              id={state.buttonId}
+              aria-haspopup="menu"
+              aria-expanded={state.isOpen}
+              data-reach-menu-button
+              type="button"
+              ref={node => {
+                rectRef(node);
+                ref && ref(node);
+                refs.button = node;
+              }}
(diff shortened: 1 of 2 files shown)


ryanflorence:
These make me nervous every time. I want to make sure people actually, eventually, render a real button there beneath all the react components. I'll have to think about this some more.

Probably breaks styled-components, doesn't it?

ryanflorence:
I'm also digging the `as` prop. `<MenuButton as={SoFancyButton}/>`

nathanforce:
@ryanflorence I'd find it bothersome to have to add a static property for that reason. That seems like a library overreaching and ultimately provides a false sense of security anyhow.

This suggestion may fly in the face of React principles, but maybe you could add all of the accessibility props via the ref rather than via a traditional props interface?  This code is using Hooks but it should get the idea across.

```
const Component = () => {
  const ref = React.useRef();
  const [isOpen, { toggle }] = useBoolean(false);
  useEffect(
    () => {
      if (ref.current) {
        ref.current.setAttribute("aria-expanded", isOpen.toString());
      }
    },
    [isOpen]
  );
  return (
    <div>
      <div
        ref={node => {
          ref.current = node;
          node.setAttribute("aria-haspopup", "menu");
          node.setAttribute("aria-expanded", isOpen.toString());
        }}
      >
        Blah
      </div>
      <Button onClick={toggle}>Open</Button>
    </div>
  );
};
```

nathanforce:
My code here is obviously adding the properties to the `div` instead of the button. This is for demonstration of the technique rather than how we'd apply it in this lib.

nathanforce:
Maybe it would be even better to use the ref to check for those aria properties on mount/change and warn in dev mode?

nathanforce:
Working demo of my idea.

https://4241xq22nw.codesandbox.io/

ryanflorence:
Thanks Nathan.

I'm thinking of just trusting people and adding this, but coming up with a "scary" prop name that makes them think twice and review the docs.

`<MenuButton accessibleButtonComponent={FancyButton}/>`

And then we just trust them that they eventually render a `<button>`, or an accessible implementation of a button (like React Native Web), and that they forward all the props, and that they use forwardRef so we can get a ref and call focus on it.

everdimension:
What about making a render prop* that passes all necessary accessibility attributes? After all, as a user of an a11y library I expect the library to help me with that.

*And when I say "render prop" you can surely think of a hook if that's more appropriate here

ryanflorence:
@everdimension the list of props is different if its a `<button>` or some other element. Also, React Native Web's `Button` element is 1000 lines of code. Wrapping all that into `<div {...accessibilityProps}/>` seems ... I dunno ... not awesome.

VinceMalone:
> These make me nervous every time. I want to make sure people actually, eventually, render a real button there beneath all the react components. I'll have to think about this some more.

I can empathize with that. It's how I feel when creating a component abstraction for styles that also accepts element props and `className`; the consumer _can_ override the styles if they didn't know better.

> I'm also digging the `as` prop.

Oh! I also dig that name.

> `asAccessibleButtonComponent`

😝 Like I said, I can empathize. Though it's worth mentioning, the issue of [an extensible API that allows the consumer to "shoot themself in the foot"] will be systemic in this library (I think.) I imagine more components will fall into this scenario (`MenuLink` already has, `MenuItem` could too, and a future component like `DisclosureSummary` probably would as well.) Will a prop name like `asAccessibleXComponent` work there too? I ask with a concern for code symmetry/consistency.

ryanflorence:
Yeah, `asAccessibleLinkComponent`, `asAccessibleButtonComponent` are probably the only two we'd need.

VinceMalone:
I've updated the prop name to `asAccessibleButtonComponent`.

everdimension:
Just a thought:

This is something that can be caught with a linter. Sort of.
Similar to how [label-has-associated-control](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md#case-my-label-and-input-components-are-custom-components) rule works and allows you to add your custom component to be interpreted as an a11y label.

Of course this solution still suffers from the problem that @nathanforce mentioned (_future (breaking) changes to the passed in component are not covered very tightly._) But so does the `asAccessibleLinkComponent` prop solution.

And the lint rule would have to come with the library, which is kind of unusual.
But perhaps it's not such a bad idea and it can become a common practice for libraries to include linting rules? 🤔

VinceMalone:
@ryanflorence any updates on the viability of merging this?

StevenLangbroek:
@ryanflorence could you maybe shine some light on what the plan is with this? Not being able to use (potentially complex) other components is a show-stopper for us. If this needs further work I'd gladly help.

StevenLangbroek:
An alternative idea is exporting the required parts to be able to reimplement this in userland (primarily missing is exporting the `Context`).

chancestrickland:
I'm going to favor closing this at moment. We will likely solve this problem by creating a separate `@reach/button` component or making sure we properly shim all necessary accessibility attributes/behaviors before we add this prop to another component. This is a priority for us since we know it's a big user need, we just need to be thoughtful about taking the right approach.

PR #108, opened 2019-01-14
[menu-button] ArrowUp after MouseLeave of non MenuItem causes runtime error

nathanforce:
Arrowing up after mousing over/out of a non Menuitem child was setting the `selectionIndex` to values that were out of range.

To reproduce on master:

1. Open Non Menu Children in storybook 
2.  Open menu
3. Arrow to an item
4.  Mouseover and out of a non MenuItem (one of the section labels)
5.  ArrowUp


chancestrickland:
I was not able to reproduce the underlying bug, so perhaps this was fixed with another recent update. @nathanforce if you continue to see this behavior, would you mind taking a short screen capture video so I can see exactly what is triggering the error?

PR #110, opened 2019-01-18
@reach/menu-button: Fix positioning and hovering issues on IE11

### LINKED ISSUE #94: Question on Browser Support
Is there a list of browsers that reach ui will plan on supporting?

I love the library and hope to use it on my current project in production however I noticed that the MenuButton dropdown does not seem to be working on my project. 

It is not working in my app or on in the demo on the website. https://ui.reach.tech/menu-button. Once you press the button the menu does not show up. It is giving me the missing styles warning probably because ie11 does not support css variables. Not sure though if that is also messing with the data-reach-menu container that is positioned absolute. 

Edge appears to be working as expected. 

Thank you.

petecorreia:
Fixes #94 

This PR fixes a couple of critical issues on IE11:

## MenuList not positioned correctly

IE11 and below don't support `window.scrollX` and `window.scrollY` which results in `top: NaNpx` and same for `left`

reach-menu-ie11-positioning-issue

The fix implemented is changing to `window.pageXOffset` and `window.pageYOffset` which are supported by all browsers. [Read more](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX#Notes)

**After fix:**

reach-menu-ie11-positioning-issue-fixed

## MenuList closes when non-selected item is hovered

IE11 and below don't support `event.relatedTarget` which results in the `onBlur` check failing incorrectly.

reach-menu-ie11-hover-issue

The fix implemented was simply to also check for `document.activeElement` too. [Read more](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement)

**After fix:**

reach-menu-ie11-hover-issue-fixed

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -413,7 +413,6 @@ let getStyles = (buttonRect, menuRect) => {
     left: `${buttonRect.left + window.pageXOffset}px`,
     top: `${buttonRect.top + buttonRect.height + window.pageYOffset}px`
   };
-  console.log("styles: ", styles);
 
   if (haventMeasuredMenuYet) {
     return {
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -348,7 +348,7 @@ let MenuListImpl = React.forwardRef(
         onBlur={event => {
           if (
             !state.closingWithClick &&
-            !refs.menu.contains(event.relatedTarget)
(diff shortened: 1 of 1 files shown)


Pau1fitz:
@ryanflorence do you think this will get merged into `master` soon? These are critical issues on IE11.

lostwun:
Any activity on this?

Issue #98, opened 2018-12-13
Prevent scroll on parent when dialog is open


sjelfull:
Could Dialog prevent scroll on body when open, or is that up to the app?

theKashey:
It should be handled by Dialog itself.
And here is one more good reason to do it - iOS Safary does not properly handle `position:fixed` elements with input. Once you will add `input`/`textarea` into Dialog, and focus them - your customer is lost.

ryanflorence:
Yeah I want this.

I've seen solutions where the screen jumps because the scroll bar gets removed (add overflow: hidden to the body).

I haven't had a chance to research other ways to do it that avoid the jump. Open to suggestions. Here's the spectrum thread on the same topic: https://spectrum.chat/reach/general/modal-dialog-should-lock-focus~3cd21c66-6c3d-4b73-adcb-10856a44eadd

ryanflorence:
(also, happy to merge a PR that simply adds `overflow: none` to `body` until then :)

theKashey:
Please don't. 
`overflow:hidden` is working mostly for desktop browsers, while problem is mostly about mobile ones. You have to silence scroll/onwheel events - it will _fix_ iOS, and prevent "normal way" of scroll on desktop - you still could drag the scroll bar, but would you?
But it's not so easy, and (today) absolutely not friendly with portals by design(killing events everywhere, except whitelisted node). While some _selects_ nowadays uses portals - that's not a great idea (anyway current `focus-trap` is also not portal-friendly).

ryanflorence:
@theKashey What is your solution then?

sjelfull:
Could https://github.com/willmcpo/body-scroll-lock be a good alternative, either by copying their approach or using it as a dependency?

theKashey:
@sjelfull - `1.1KB` total size seems quite good, twice better than "react" locks. And `allowTouchMove` callback probably could be used to handle portals in the future.

@ryanflorence - mine solution nowadays would block any interactions with _portaled_ content, ie `menu`, and twice bigger than body-lock.

karlhorky:
#111 has an issue with scroll locking: reach/reach-ui#161 (due to theKashey/react-remove-scroll#4).

donaldpipowitch:
[...]

Issue #83, opened 2018-10-15
dialog: focus-trap blocks interactive elements rendered using portal


kirillku:
Hi, thanks for the `Dialog`. It works great but I face one issue. You are using `focus-trap` internally to block events outside the `Dialog`, which is nice and I understand the logic behind this. I am rendering some interactive elements inside the `Dialog` and some of them are also using portals for rendering (for example popups), which results them being not a child node of the `Dialog` and this why not active. They are visible, but not clickable. Can you suggest some solution or workaround? I didn't come up with anything. Maybe it's possible to only block click events on the elements, that are above the `Dialog` in the DOM tree.

everdimension:
Interesting case.

I think the right solution is to make your portal elements render inside the dialog component. It actually makes sense.

The reason for rendering tooltips and popups as a direct child of body is usually either dealing with css specificity or z-index wars. And since you have a modal dialog open, you're already rendering your elements outside the main app root. There should be no need to render popups in a parallel branch.

kirillku:
Yep, this is what I would do with my own components to solve the issue, but what about third party ones. For example [react-semantic-ui Popup](http://react.semantic-ui.com/modules/popup/).

Also, the reasoning behind putting something in a portal can be not only z-index, but also overflow positioning.

everdimension:
> Yep, this is what I would do with my own components to solve the issue, but what about third party ones. For example react-semantic-ui Popup.

I had similar problems with 3rd party libraries and I've came to the conclusion that those libs that do not let you configure the root container they attach to are not very good libraries :) And it is a good idea to open an issue in their repository.

I understand that this is not the answer you were hoping for, but it seems to me to be the "right" way to deal with this.

---

Also I wonder whether the `@reach/dialog` allows you to specify a root other than the body 🤔  
Seems to me that it doesn't https://github.com/reach/reach-ui/blob/master/packages/portal/src/index.js#L10 🙁

kirillku:
What about making `disableFocusTrap` prop for such cases. I think there is a lot 3rd party libs, that only attach portals as direct children of body. I will do a mr for that.

theKashey:
[react-focus-lock](https://github.com/theKashey/react-focus-lock) could be a solution here - it uses React event propagation, not DOM, thus focus events, as long as they would _bubble_ from portals, are visible to lock, and are considered as their own.
But first we have to remove focus-trap from reach/dialog.

ryanflorence:
[...]

PR #111, opened 2019-01-22
Add react-focus-lock and react-remove-scroll

theKashey:
React-Focus-Lock fixes:
- #83, it accepts _focus_ events via React Event System, thus count portaled content as own content.
- unreported issue about nested dialogs. Once second dialog is closed lock is no more active.
- `initialFocusRef` was not drilled to Dialog

Scroll-Lock(react-remove-scroll) fixes:
- #98, `position:fixed` elements have to disable body scroll on iOS
- unreported issue, alerts and modals should lock scroll(by design).
- also support portals, nested locks and so on

+ a bit simpler code about focus management
+ a few examples to prove/test new stuff

### DIFF
--- a/packages/dialog/examples/autofocus.example.js
+++ b/packages/dialog/examples/autofocus.example.js
@@ -1,0 +1,28 @@
+import React from "react";
+import Component from "@reach/component-component";
+import "../styles.css";
+import { Dialog } from "../src/index";
+
+export let name = "Autofocus";
+
+export let Example = () => (
+  <Component
+    initialState={{ showDialog: false }}
+    refs={{ button: React.createRef() }}
+  >
+    {({ state, setState, refs }) => (
+      <div>
+        <button onClick={() => setState({ showDialog: true })}>
+          Show Dialog
+        </button>
+
+        <Dialog isOpen={state.showDialog} initialFocusRef={refs.button}>
+          <button onClick={() => setState({ showDialog: false })}>
+            Close Dialog
+          </button>
+          <button ref={refs.button}>Auto focused</button>
+        </Dialog>
+      </div>
+    )}
+  </Component>
+);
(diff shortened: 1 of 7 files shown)


packages/dialog/examples/long-content.example.js
```diff
@@ -13,6 +13,16 @@ export let Example = () => (
           Show Dialog
         </button>
 
+        {new Array(20)
+          .fill(1)
+          .map((x, index) => (
+            <div
+              style={{ position: "absolute", top: (index + 1) * 100 + "px" }}
+            >
+              - scroll -
```
theKashey: Just to test scroll lock.

ryanflorence:
merging and releasing this today, thanks for your patience and for the code :D

ryanflorence:
done, @reach/dialog@0.2.0

Thanks again!

theKashey:
Happy to reach reach. From now a part of your issues are mine :)

karlhorky:
I guess this causes the issue that I'm experiencing with the scroll locking not working: reach/reach-ui#161 (upstream theKashey/react-remove-scroll#4).

mkarajohn:
>unreported issue about nested dialogs. Once second dialog is closed lock is no more active. 

Was it [this](https://github.com/reach/reach-ui/issues/130) one? If that's the case I would have loved it if the issue had been mentioned here so I would have been notified that this was fixed.

theKashey:
@mkarajohn - yeah, that was it.

mkarajohn:
Huh, that's a bummer. Why was it considered an "unreported issue"?

theKashey:
Cos you added it __months__ after I've created this PR? And a few weeks before it got merged :)

mkarajohn:
Hah, holy shit, you are right. For some reason the merge date registered as the opening date in my brain, no kidding. Way to make an ass out of myself 👏

PR #115, opened 2019-01-31
Fix getFocusableMenuChildren receives non array children

### LINKED ISSUE #114: MenuList with one MenuItem throws error
If I use MenuList with one MenuItem, it crashes unexpectedly 😱 

Here is a simple demonstration (with codesandbox link: https://codesandbox.io/s/wo3j9xl8k8)

```Javascript
import React from "react";
import ReactDOM from "react-dom";

import {
  Menu,
  MenuList,
  MenuButton,
  MenuItem,
  MenuLink
} from "@reach/menu-button";

import "./styles.css";

function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <Menu>
        <MenuButton>
          Actions <span aria-hidden>▾</span>
        </MenuButton>
        <MenuList>
          <MenuItem onSelect={() => alert("Download")}>Download</MenuItem>
        </MenuList>
      </Menu>
    </div>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

Error message:
```
Uncaught TypeError: children.filter is not a function
    at getFocusableMenuChildren
```

// edit: It work's with version 1.0.7
It looks like the error comes from here because children is somehow not defined -> https://github.com/reach/reach-ui/commit/1a245ecec0fe93ff4d49bb30d9e1b48c27464fcd#diff-13ae4985661da7cf08440d52edf169faR320

The error is quite simple. `<Foo><div>You ok?</div></Foo>` does not give us an array as children - single object. `React.Children.toArray(children)` might help?

wiesson:
This PR attempts to fix #114

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -327,9 +327,7 @@ let focusableChildrenTypes = [MenuItem, MenuLink];
 
 let isFocusableChildType = child => focusableChildrenTypes.includes(child.type);
 let getFocusableMenuChildren = children => {
-  return React.Children.toArray(children).filter(child =>
-    isFocusableChildType(child)
-  );
+  return [].concat(children).filter(child => isFocusableChildType(child));
 };
 
 let MenuListImpl = React.forwardRef(
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -327,7 +327,9 @@ let focusableChildrenTypes = [MenuItem, MenuLink];
 
 let isFocusableChildType = child => focusableChildrenTypes.includes(child.type);
 let getFocusableMenuChildren = children => {
-  return children.filter(child => isFocusableChildType(child));
+  return React.Children.toArray(children).filter(child =>
+    isFocusableChildType(child)
+  );
 };
 
 let MenuListImpl = React.forwardRef(


nathanforce:
We need to be careful here as I don't think `Children.toArray` maintains identity with `Children.map`.  I think this will break here https://github.com/reach/reach-ui/pull/115/files#diff-13ae4985661da7cf08440d52edf169faR379.

wiesson:
Hmm, you are probably right. How about `[].concat(children).filter(..)` ?

nathanforce:
In our fork I just lifted the children array up.

https://github.com/palmerhq/reach-ui/commit/54a71575c4e07e6e20e0da7b41c3771c003f3e94#diff-13ae4985661da7cf08440d52edf169faR339

wiesson:
Okay, so are you going to submit your PR that fixes the issue?

nathanforce:
@ryanflorence what do you think? Instinct tells me you're in the midst of a TS/hooks rewrite so I'm hesitant to touch anything. That being said, there are some outright bugs with Menu and the handling of `children` from my PR that I feel some obligation to address.

wiesson:
If your PR fixes the issue and if there is a ongoing rewrite with hooks/ts, I'll close my PR!

PR #127, opened 2019-03-21
Support ref objects as refs for dialog components

### LINKED ISSUE #126: Uncaught TypeError: forwardRef is not a function
Trying to add a ref to `DialogOverlay`, and getting this error. We're on React 16.5.

image

Looks like `forwardRef` now returns a object instead of a function, and on this line your using it as a function: https://github.com/reach/reach-ui/blob/master/packages/dialog/src/index.js#L98

Would it be correct to change it into something like:
```javascript
ref: function ref(node) {
    if (forwardRef && forwardRef.current !== null) {
        return forwardRef;
    }
}
```

How could I make it easy for you to tag a new release that fixes this? It's currently a blocking issue for us.

Andarist:
I haven't yet tested this thoroughly, so keep that in mind. Wanted to get the ball rolling to see if you like the used approach.

fixes #126

### DIFF
--- a/packages/dialog/package.json
+++ b/packages/dialog/package.json
@@ -15,7 +15,8 @@
     "@reach/component-component": "^0.1.3",
     "@reach/portal": "^0.1.3",
     "@reach/utils": "^0.2.0",
-    "focus-trap": "^3.0.0"
+    "focus-trap": "^3.0.0",
+    "use-composed-ref": "^1.0.0"
   },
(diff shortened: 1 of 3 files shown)


Andarist:
@ryanflorence friendly 🏓

chancestrickland:
Sorry this had to sit for so long! This should be resolved in another update, so stay tuned for the next release.

PR #137, opened 2019-04-09
Append portaled content to the correct (owner) document

bvaughn:
This afternoon I tried plugging the `MenuButton` component into the new React DevTools (https://github.com/bvaughn/react-devtools-experimental/pull/108) but I ran into a snag. The DevTools extension portals itself into Chrome panels, but the Reach `Portal` component assumes its content should be appended to the root/original `document`. The result is that the `MenuButton` portal (and others) are essentially invisible.

This PR fixes that. Posting for discussion purposes! 🙇

cc @ryanflorence

### DIFF
--- a/packages/portal/src/index.js
+++ b/packages/portal/src/index.js
@@ -4,17 +4,25 @@ import Component from "@reach/component-component";
 
 let Portal = ({ children, type = "reach-portal" }) => (
   <Component
-    getRefs={() => ({ node: null })}
+    getRefs={() => ({ mountNode: null, portalNode: null })}
     didMount={({ refs, forceUpdate }) => {
-      refs.node = document.createElement(type);
-      document.body.appendChild(refs.node);
+      // It's possible that the content we are portal has, itself, been portaled.
+      // In that case, it's important to append to the correct document element.
+      const ownerDocument = refs.mountNode.ownerDocument;
+      refs.portalNode = ownerDocument.createElement(type);
+      ownerDocument.body.appendChild(refs.portalNode);
       forceUpdate();
     }}
-    willUnmount={({ refs: { node } }) => {
-      document.body.removeChild(node);
+    willUnmount={({ refs: { portalNode } }) => {
+      portalNode.ownerDocument.body.removeChild(portalNode);
     }}
-    render={({ refs: { node } }) => {
-      return node ? createPortal(children, node) : null;
+    render={({ refs }) => {
+      const { portalNode } = refs;
+      if (!portalNode) {
+        return <div ref={div => (refs.mountNode = div)} />;
+      } else {
+        return createPortal(children, portalNode);
+      }
     }}
   />
 );


packages/portal/src/index.js
```diff
@@ -4,17 +4,25 @@ import Component from "@reach/component-component";
...
     }}
-    willUnmount={({ refs: { node } }) => {
-      document.body.removeChild(node);
+    willUnmount={({ refs: { portalNode } }) => {
+      portalNode.ownerDocument.body.removeChild(portalNode);
     }}
-    render={({ refs: { node } }) => {
-      return node ? createPortal(children, node) : null;
+    render={({ refs }) => {
+      const { portalNode } = refs;
```
bvaughn: We could use the same ref for both, but I though this was maybe cleaner?

packages/portal/src/index.js
```diff
@@ -4,17 +4,25 @@ import Component from "@reach/component-component";
...
 let Portal = ({ children, type = "reach-portal" }) => (
   <Component
-    getRefs={() => ({ node: null })}
+    getRefs={() => ({ mountNode: null, portalNode: null })}
     didMount={({ refs, forceUpdate }) => {
-      refs.node = document.createElement(type);
-      document.body.appendChild(refs.node);
+      // It's possible that the content we are portal has, itself, been portaled.
+      // In that case, it's important to append to the correct document element.
+      const ownerDocument = refs.mountNode.ownerDocument;
```
bvaughn: Maybe there's a way for a component to get the `ownerDocument` without going through the creation of a ref. I'll have to dig into ReactDOM's source tomorrow and see.

bvaughn:
FWIW I've [added a forked build of `@react/portal` to the new React DevTools](https://github.com/bvaughn/react-devtools-experimental/tree/master/vendor/%40reach/portal) with this fix and it's working well for us so far 😄

packages/portal/src/index.js
giuseppeg: from your experience do you think that it is better to append to the body or is this approach fine https://github.com/giuseppeg/react-layers-manager/blob/master/src/LayersManager.js#L13-L16 ?

bvaughn: TBH, I'm not familiar with rhe isolation property. Looks like it's not supported by IE/Edge though so likely a deal breaker there.

giuseppeg: oh I didn't mean that exact piece of code, but yea isolation creates a stacking context without affecting layout and the provider is just wrapping the app to trap any stacking context in the app. Then it creates a separate one where it appends the <Layer> (portals) instances. I'll fix the public repo to support ie/edge

I was wondering if, because of other 3rd parties libs, it'd be better to instead append to the body all the time

bvaughn: I don't know. I haven't done a lot of work with portals

PR #144, opened 2019-04-21
chore: Upgrade to using terser

maraisr:
With this PR I aim to use `terser` instead of uglifyjs since it doesn't support es6.


chancestrickland:
@maraisr Looks like with time we've got a few minor conflicts. Mind cleaning those up? Happy to merge this one if so.

maraisr:
Hi - I've deleted my fork, and thus the reference branches, and because of that I can no longer make changes to this branch. However, I do believe you as a repo owner have write access to this PR.

Maybe a simple, `git checkout origin/master -- yarn.lock; git rebase origin/master` will do the trick. :)

PR #148, opened 2019-04-23
Lift menu list class name to outer div

brendancarney:
**Note:** Breaking Change

Because the `data-reach-menu` element is rendered in a portal, there's no way (that I'm aware of) to scope styles specifically to one menu. I can style `data-reach-menu`, but only globally. 

Screen Shot 2019-04-22 at 8 38 17 PM

Maybe I'm missing something there? 

What this does is lift the `className` from `MenuList` to that outer div. This does present some naming issues, as it seems odd to have the class name from `MenuList` style the `data-reach-menu` element. 

Any thoughts?

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -287,7 +287,7 @@ MenuLink.propTypes = {
 };
 ///////////////////////////////////////////////////////////////////
 
-let MenuList = React.forwardRef((props, ref) => (
+let MenuList = React.forwardRef(({ className, ...props }, ref) => (
   <Consumer>
     {({ refs, state, setState }) =>
       state.isOpen && (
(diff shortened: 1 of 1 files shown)


brendancarney:
Just today I finally reached the point where I actually have to style two menu lists differently, and I'm struggling to find a workaround. @ryanflorence any thoughts on this? Thank you for your time!

chancestrickland:
Closing in favor of https://github.com/reach/reach-ui/pull/310 so we can introduce more flexibility without a breaking change. Seems a little more intuitive as well to have the outer menu represent the `div` inside the portal, even though the structure is technically different than the HTML output (Menu component props passed to data-reach-menu makes sense)

chancestrickland:
After further discussion, I reverted #310, but I do want to explore limitations of the current API to see if we might need a better approach. I'll open a separate issue for this and we'll continue the conversation there.

brendancarney:
The use case is adding padding to the wrapper and setting a background on it so that I can have a scroll bar that looks like this:
Screen Shot 2019-10-16 at 1 59 07 PM

I don't think this would be possible without the wrapper, but I'm not 100% sure. 

Maybe someday there could be a lower level implementation to hook into instead, so that odd use cases like this could build their own tree? Some `useMenu()` hook or something. 

I did end up with a pretty hacky workaround, which is to modify the styles of that node whenever the menu is opened.

PR #150, opened 2019-04-25
Publish menu-button type definitions

hebbeh:
The type definitions added in #92 were not published with the npm package. 😄

### DIFF
--- a/packages/menu-button/package.json
+++ b/packages/menu-button/package.json
@@ -5,6 +5,7 @@
   "author": "Ryan Florence <@ryanflorence>",
   "license": "MIT",
   "main": "index.js",
+  "types": "index.d.ts",
   "module": "es/index.js",
   "scripts": {
     "test": "echo \"Write some tests you bum!\"",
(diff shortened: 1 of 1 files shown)


packages/menu-button/package.json
```diff
@@ -5,6 +5,7 @@
   "author": "Ryan Florence <@ryanflorence>",
   "license": "MIT",
   "main": "index.js",
+  "types": "index.d.ts",
```
disintegrator: I think this line is not required because typescript will, by default, detect an index.d.ts file at the root of a package

focux:
When can we expect this to be merged? I'm facing the same issue.

packages/menu-button/package.json
pirelenito: You are correct that it is not needed, however the TypeScript documentation still recommends to be explicit:

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so

From: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

PR #154, opened 2019-04-29
Account for possible scrollbars in the Tooltip triangle sample

petetnt:
Okay, this might seem kinda dumb but a common thing with tooltips in my experience: on platforms where scrollbars are visible the tooltip causes scrollbars to appear if the tooltip is too close to the said scrollbar:

image

Deducting the [common max of scrollbar widths](https://codepen.io/sambible/post/browser-scrollbar-widths) from `maxLeft` removes the issue:

image

And the tooltips that don't touch the scrollbar still work as expected:

image


chancestrickland:
@petetnt Thanks for flagging! I made a few tweaks to calculate the width of the scrollbar only when its visible and it looks like this should be fixed. 👍

petetnt:
💃 thanks!

PR #155, opened 2019-04-30
Fixed URL for lerna

hriziya:
Fixed URL for lerna


evocateur:
Thanks for noticing this. :)

Issue #160, opened 2019-05-03
❓ tooltip: should wrap onBlur instead onFocus?


rexpan:
Is this a mistake or intention for wrap `onBlur` with `onFocus`? 
https://github.com/reach/reach-ui/blob/99c687b6d9eca174133e0f5beadafd1842be2214/packages/tooltip/src/index.js#L343

chancestrickland:
Fixed in https://github.com/reach/reach-ui/pull/226

PR #162, opened 2019-05-05
Fix onBlur handling

asistapl:


asistapl:
Fixes #160

chancestrickland:
Fixed in https://github.com/reach/reach-ui/pull/226

PR #164, opened 2019-05-06
Upgrade react-remove-scroll to 1.0.8

### LINKED ISSUE #161: Scroll not locked when moving mouse out of window after scroll (Chrome, macOS)
#111 added scroll locking to the background elements.

It works well except for the following case (admittedly, an edge case):

1. Scroll a lot with a mouse/setup that allows acceleration scrolling (ex. I have a Magic Mouse)
2. Move mouse outside of window quickly afterwards
3. Watch the locked background scroll with the left over acceleration

Kapture 2019-05-03 at 13 13 43

I have also added an issue upstream: theKashey/react-remove-scroll#4

karlhorky:
Fixes #161

### DIFF
--- a/packages/dialog/package.json
+++ b/packages/dialog/package.json
@@ -16,7 +16,7 @@
     "@reach/portal": "^0.2.1",
     "@reach/utils": "^0.2.2",
     "react-focus-lock": "^1.17.7",
-    "react-remove-scroll": "^1.0.2"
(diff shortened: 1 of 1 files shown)


chancestrickland:
Sorry for the delay here but thank you for contributing! This was updated in another similar PR in the backlog.

PR #172, opened 2019-05-14
@reach/rect: Resubscribe when nodeRef changes

einarlove:
Before this change, if you passed a new ref or changed the `ref.current` passed to `useRect` then you would never get rects from the new node. This update fixes this by resubscribing the listener if the node changes.

This is only a fix for the useRect hook, not the 2018-render-props thingy. It's incredible how easy reasoning with updates are with hooks vs render-props.

I'll take a look into fixing the `<Rect>` as well if you approve of this change.


einarlove:
After checking the docs, my conclusion is that this update can't be applied to `<Rect>` since it can't know when/if the ref changes. The only onUpdate is triggered by `observe` change. So for this update to work for `<Rect>`, some refactoring has to be done.

einarlove:
I added a commit which checks for nodeRef.current before subscribing so that if you pass a ref which does not _yet_ have a node reference it silently waits.

```js
const [anchor, setAnchor] = useState()
const anchorRect = useRect({ current: anchor })

return (
  <div>
    {JSON.stringify(anchorRect)}
    {items.map(item =>
      <Item onClick={event => setAnchor(event.target)} />
    )}
  </div>
)
```

mjackson:
LGTM. We should update the `<Rect>` API too though, so they match. Feel free to not use the `<Component>` API and just use hooks instead.

mjackson:
After thinking about this a bit more, I don't think this is safe. The problem is we can't observe a mutable value, so adding it to the dependency array doesn't make sense.

I believe we need to actually return a ref from `useRect` instead of taking a ref as an argument.

```js
// instead of
let ref = useRef()
let rect = useRect(nodeRef)

// we should do
let [rect, ref] = useRect()
```

And then let users apply the ref wherever they need it. If we are the ones creating the ref, we can know when it changes by using a function ref. Otherwise we can't know.

I'd recommend we revert this change and revisit in another PR.

einarlove:
I disagree wholeheartedly. Hooks should not supply their own refs, they should accept one. How should I solve this?
```js
const ref = React.useRef()
const rect = useRect(ref)
onClickOutside(ref, () => props.close())
focusTrap(ref)

return (
  <div ref={ref}>
    …
```
unless you expect the users to either add a `div` for each hooks that wants a ref or use convoluted functions for merging refs?

mjackson:
Ya, agree our hook should not return a ref. Still, adding a ref to the dependency array of `useLayoutEffect` breaks things because we can't observe when it changes.

One possible solution would be to add a second config argument to `useRect`, like a `key` that would tell us when your ref changes. Maybe something like

```js
let rect = useRect(ref, { observe: true, key: key })
```

Then at least we could know when to re-run the layout effect. Another solution (although kind of terrible) would be to just re-render the owner of `useRect` with a different `key`, so you'd get a new instance of the hook.

But for our current purposes `useRect` works just fine and this change breaks it, so I'm going to revert for now. Feel free to propose another PR with the API I suggested above, or just copy our `observeRect` function and make your own hook with it.

chancestrickland:
@einarlove As @mjackson mentioned, this change did have some undesirable breaking changes. If you run can reproduce a this issue with an example I'd be happy to take another look at it.

Issue #174, opened 2019-05-16
@reach/tooltip IE 11 is not supported


andrew-humu:
## Overview

Tooltips don't appear to be visible in any case in IE11.

It looks like the only issue is that IE11 doesn't support `window.scrollX` or `window.scrollY`.

https://developer.mozilla.org/en-US/docs/Web/API/window/scrollX

## Version tested

`"@reach/tooltip": "0.2.0"`

ryanflorence:
thanks! fixed and released.

andrew-humu:
Amazing, thank you!

PR #175, opened 2019-05-16
Use pageOffset instead of scrollX and scrollY

andrew-humu:


andrew-humu:
IE11 doesn't support `window.scrollX` or `window.scrollY`.

https://developer.mozilla.org/en-US/docs/Web/API/window/scrollX#Browser_compatibility

But IE11 does support `window.pageXOffset` and `window.pageYOffset`.  MDN claims this is an exact alias.

https://developer.mozilla.org/en-US/docs/Web/API/window/pageXOffset#Browser_compatibility

Fixes #174

chasemccoy:
@ryanflorence We can't use `@reach/tooltip` until it works in IE, unfortunately 😢 Would love to see this get merged!

PR #176, opened 2019-05-20
Simplify Tooltip statechat management

Andarist:
By definition statecharts ignore events for which they don't have defined transitions on the active state node. So it seems to me this is counterproductive to guard against "undefined" transitions at the event dispatch site. Main reason for that is that whenever you want to tweak your statechart definition you'd have to tweak your transition call sites too


Andarist:
@ryanflorence friendly 🏓 any thoughts on this?

chancestrickland:
Closing this along with other old PRs for now just to do some housekeeping. We can reopen in the future if we feel like this is needed.

PR #183, opened 2019-05-23
Fix missing Auto ID page on documentation website

lukyth:
Right now there's an `Auto ID` menu at https://ui.reach.tech/, which will lead to https://ui.reach.tech/auto-id, and show the `NOT FOUND` page.

This PR adds a page for `Auto ID` (`@reach/auto-id`) and fix that issue.

### DIFF
--- a/www/src/pages/auto-id.mdx
+++ b/www/src/pages/auto-id.mdx
@@ -1,0 +1,31 @@
+import Helmet from "react-helmet"
+
+<Helmet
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Thanks for the fix. Is this supposed to be deployed yet? I’m still getting a 404.

chancestrickland:
@nickmccurdy I'm new here, had to get the proper credentials to deploy. Should be all set! 👍

PR #189, opened 2019-06-03
Handle left collision with a wide MenuList

mrm007:
`@reach/menu-button` has an odd behaviour when the `MenuButton` and/or `MenuList` are wide. See the bug in action:

<details>
<summary>GIF</summary>

patched

</details>

In (other) words, when the list is wider than the button the layout algorithm always assumes that a collision with the left side has occured. This works fine for left aligned menus, but not so much for right aligned ones. Since the list is not wider than the viewport there is enough space to expand it to the left.

Sandboxes are worth 1000 words, so I created [this one][sandbox] to play around in.

[sandbox]: https://codesandbox.io/s/misty-darkness-zjepf


mrm007:
@ryanflorence any chance you can take a look at this? Let me know if this thing needs anything more to get it merged.

PR #195, opened 2019-06-08
Fix border styles of <ComboBox/> pink demo

Andarist:
Without this, if we land focus outside of the browser tab instead of on another element when bluring then we'll see such broken styling:

Screen Shot 2019-06-08 at 20 45 54

### DIFF
--- a/www/src/components/app.css
+++ b/www/src/components/app.css
@@ -285,7 +285,8 @@ button#hamburger {
 }
 
 .pink [data-reach-combobox-popover] {
-  border: none;
+  border: solid 1px #ccc;
+  border-top: none;
   box-sizing: border-box;
   position: relative;
   top: -1px;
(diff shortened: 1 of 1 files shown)


chancestrickland:
@Andarist I'm not able to reproduce this issue. Would you be able to take a screen recording so I can see the exact behavior that leads to this effect?

Andarist:
This demo suffers from it - https://ui.reach.tech/combobox#custom-styling . Repro steps:
1. open browser dev tools
2. type something into that combobox (from the custom styling demo), e.g. "new"
3. while combobox is opened click on the devtools

Please verify if my fix is actually fixing this right now, this was quite a long time ago - so things could have changed in the meantime.

chancestrickland:
Closing this now as it will be fixed upstream in the next deployment. Unfortunately it just sat a bit long and got stale with other changes, but I do appreciate the call-out!

PR #198, opened 2019-06-09
Fix broken memo (recreated every render) in ComboboxOptionText

Andarist:


packages/combobox/src/index.js
```diff
@@ -576,12 +576,14 @@ export function ComboboxOptionText() {
     data: { value: contextValue }
   } = useContext(Context);
 
-  const searchWords = escapeRegexp(contextValue).split(/\s+/);
```
Andarist: this has yielded new reference each time

PR #200, opened 2019-06-10
(docs) Fix Tabs anchors

aarongarciah:


chancestrickland:
Thank you for submitting! We had a few PRs open that addressed the same issue. This should be resolved. 👍

Issue #166, opened 2019-05-07
Tooltip bonanza bug


bvaughn:
Noticed this issue testing the React profiler:
![@reach-tooltip-bug-Kapture 2019-05-07 at 14 52 16](https://user-images.githubusercontent.com/29597/57335776-8f3e9180-70d8-11e9-8a39-a73462e47b23.gif)

Can be reproduced by going [here](https://react-devtools-experimental.now.sh/) and importing [this `profiling.json` export](https://gist.github.com/bvaughn/b6946712957bdcdc3c031ded59bb3c72).

I am working on a smaller repro but haven't yet figured it out. 😅 

Note that this bug occurs in tooltip versions 0.1 and 0.2.

bvaughn:
Here's a smaller repro:
https://qkyvl35j09.codesandbox.io/

paulconlin:
I've seen this issue a few times in my project as well, though I can figure out the steps to reproduce it as consistently as your example ^

FlipgridMartin:
@bvaughn I noticed that the tooltip bug doesn't happen if you remove the Tooltip from the initial `repro` button. Possibly a timing subscriptions issue?

https://codesandbox.io/s/reachtooltip-bonanza-bug-repro-k8fwg

bvaughn:
Fair enough, although that's not really an option in my project (DevTools) since that button needs a tooltip as well :)

FlipgridMartin:
Totally understand. I was just hinting towards a possible cause for the issue, our team is in the same boat!

SavePointSam:
I found a possible solution. Though, I want to make sure that it does not have an unintended side effects before I post a PR.

SavePointSam:
@bvaughn, I was able to correct all the tooltips showing upon transition with my WIP solution. However, your demo does shift focus away from and back to the button that was originally clicked (default browser behavior, AFAIK). I added a red outline to the focused element in the GIF below demoing my solution. Is this your desired behavior?

![2019-06-11 16 08 50](https://user-images.githubusercontent.com/8203211/59312947-c407b080-8c63-11e9-8938-9768bc3c911a.gif)

bvaughn:
I'm not really sure what you're asking. I don't explicitly set focus, so... it seems like if the browser's default behavior causes tooltips to break, that's not good. (There are a few cases where this happens in DevTools. This was just one repro I was able to isolate.)

SavePointSam:
[...]

PR #203, opened 2019-06-12
prevent all tooltips from showing initially based on global state

SavePointSam:
Given #166, there is an issue of new Tooltips being shown by default if the current global state happens to be true for any other Tooltip (as there is a single state container for ALL Tooltips in an application).

This patch addresses that problem by only recovering the global state if the ID of the new usage of `useTooltip` has the same ID as the currently active ID.

However, #166's use case does still present an interesting issue, even after this patch. Having the default state be based on the ID of the Tooltip can lead to false positives as IDs are reused within the life of the the Application. Meaning that Tooltips that aren't actually meant to be shown as active appear as active. Although, this may be better to file as a separate issue.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -10,7 +10,7 @@
 //
 // Only one tooltip can be visible at a time, so we use a global state chart to
 // describe the various states and transitions between states that are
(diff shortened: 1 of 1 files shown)


ryanflorence:
published in 0.2.2

Issue #158, opened 2019-05-02
Tooltip:  arria attributes do not have valid values


moonmeister:
Lighthouse is complaining about invalid aria values on tooltip.

[aria-*] attributes do not have valid values

```html
<svg class="icon-module--icon--27QhR" viewBox="0 0 24 24" aria-labelledby="tooltip:1" data-reach-tooltip-trigger="">
  <path d="the long path is here"></path>
</svg>
```

Thanks for all the work on this, this made the tooltip project I've been putting off real quick.

chancestrickland:
This should be fixed with https://github.com/reach/reach-ui/commit/85e41a91ba9b46f6231b1d419af61cd70f1e9941

PR #207, opened 2019-06-14
[Tooltip] Fix invalid aria-describedby attribute 

darellkoh:
## Issue
After implementing tooltips, I ran into the same Lighthouse `[aria-*] attributes do not have valid values` audit warnings referenced in [this issue](https://github.com/reach/reach-ui/issues/158), but for the `aria-describedby` attribute.

Lighthouse audit:

Screen Shot 2019-06-14 at 12 43 38 AM

Affected component:

Screen Shot 2019-06-14 at 12 51 13 AM

## Fix
The `aria-describedby` attribute is considered invalid if the tooltip is not visible, since the id referenced does not yet exist on the page. By showing the `aria-describedby` attribute only when the tooltip is visible, we will ensure that the attribute is valid. 

This fixed the Lighthouse warning for my tooltip implementation and raised the overall page accessibility score from 84 to 86.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -333,8 +333,7 @@ export function useTooltip({
     }
   };
 
-  const trigger = {
-    "aria-describedby": id,
+  let trigger = {
     "data-reach-tooltip-trigger": "",
     ref: triggerRef,
     onMouseEnter: wrapEvent(onMouseEnter, handleMouseEnter),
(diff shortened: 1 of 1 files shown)


chancestrickland:
Thanks for the PR! Sorry we didn't review this sooner, but the issue should be resolved with https://github.com/reach/reach-ui/commit/85e41a91ba9b46f6231b1d419af61cd70f1e9941.

PR #209, opened 2019-06-14
useLayoutEffect deps

nathanforce:
Not sure if this is an actual issue, but https://github.com/kentcdodds/stop-runaway-react-effects is throwing for these as they run on each render.

I figure it doesn't hurt to have them only fire when needed?


nathanforce:
To be clear, the root issue was my own looping render, but it surfaced the missing deps here.

PR #220, opened 2019-06-25
Make @reach/dialog responsive

mxstbr:
50vw on mobile and tablet screens is not great, dialogs should be way bigger there. This switches from 90vw to 50vw on devices bigger than tablets (at least according to the media queries in `skeleton.css` for the website)

### DIFF
--- a/packages/dialog/styles.css
+++ b/packages/dialog/styles.css
@@ -23,7 +23,7 @@
   outline: none;
 }
 
-@media (screen and min-width: 750px) {
+@media screen and (min-width: 750px) {
   [data-reach-dialog-content] {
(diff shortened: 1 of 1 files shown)


mxstbr:
I feel like on mobile it should be full-screen, to be honest. What do you reckon @ryanflorence?

chancestrickland:
This is something you should be able to handle fairly easily in your app, as opinions will differ as to what is expected depending on what content you use inside the dialog. We provide some basic default styling as a convenience so that it functions off the shelf, but we're probably going to keep those styles as simple as possible and give you the tools to customize as you see fit.

Issue #216, opened 2019-06-19
[Combobox] duplicate values prevent arrow navigation


aweary:
## Spark Notes Summary

If there are two `ComboboxOption`s with the same `value` it prevents you from navigating past them with the arrow keys.

## Prove It

* Go to [this Codesandbox](https://codesandbox.io/s/reach-combobox-duplicate-value-arrow-bug-si10y)
* Focus the input and type "A"
* Focus the first "Apple" result by pressing the down arrow key
* Attempt to focus the next result by pressing the down arrow key again

## What Happens?

Focus does not move. You can kind of work around this by pressing *up* and wrapping around the whole list but that sucks

## What Should Happen?

I don't really know, maybe it should skip the next "Apple" value and select the next unique option after that? Having duplicate values is generally a bad experience since it can move focus and trap you in this weird focus cycles. Maybe the answer is just "don't have duplicate values"?

chrstntdd:
I hit this bug as well when messing around with the combobox, but part of me thinks that it should definitely be the responsibility of the dev to filter out duplicate values before rendering them so as not to confuse end users with multiple seemingly identical - yet distinct choices. 

Since the combobox active-descendant is not being derived from the item's index - this seems unfeasible given the design decisions Ryan made [here](https://github.com/reach/reach-ui/blob/master/packages/combobox/src/index.js#L693).

aweary:
I think it would be solvable by using a combination of the value and index to uniquely identify options, and then using `findIndex` to calculate the index instead of `indexOf`.

It's totally up to the product developer to provide unique values but I'm 99% sure many people will accidentally have duplicate values _eventually_ and it'd be cool if it didn't result in super confusing behavior for the user. Maybe checking for duplicates in DEV and using `React.error` or React.warn` to call them out would be a nice way to prevent that.

chrstntdd:
That's a great idea! If @ryanflorence is into it - I'd love to contribute!

fnky:
I don't think this should be the job of the Combobox itself. In my opinion it should work like the `key` prop and not allow duplicate values by throwing an error that explains this in detail.

PR #222, opened 2019-06-28
[Combobox] Add index to uniquely identify duplicate options.

chrstntdd:
Devs shouldn't be providing duplicate options for users to choose from, but with this we can at least keep trucking along without running into the issue described in #216

With this change, we could also remove the need for running the hashing function to track the active descendant, but I'll leave that up to you.

### DIFF
--- a/packages/combobox/src/index.js
+++ b/packages/combobox/src/index.js
@@ -166,7 +166,7 @@ function reducer(data, action) {
     case SELECT_WITH_KEYBOARD:
       return {
         ...nextState,
-        value: data.navigationValue,
+        value: stripIndex(data.navigationValue),
         navigationValue: null
       };
     case CLOSE_WITH_BUTTON:
--- a/packages/combobox/src/index.js
+++ b/packages/combobox/src/index.js
@@ -201,6 +201,8 @@ const findNavigationValue = (state, action) => {
   }
 };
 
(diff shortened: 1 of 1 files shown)


packages/combobox/src/index.js
```diff
@@ -398,7 +400,9 @@ export const ComboboxInput = forwardRef(function ComboboxInput(
   const inputValue =
     autocomplete && (state === NAVIGATING || state === INTERACTING)
       ? // When idle, we don't have a navigationValue on ArrowUp/Down
-        navigationValue || controlledValue || value
+        (navigationValue && stripIndex(navigationValue)) ||
```
chrstntdd: Heck, this function can just be inlined.

chrstntdd:
This is also currently missing a call to `stripIndex` at the `SELECT_WITH_KEYBOARD` action.

chrstntdd:
If this ends up being merged in - it would seem that the hashing function could be removed. Let me know!

chancestrickland:
@ryanflorence If you have time to glance at this quickly, can you let me know if you see any flags? I tested this locally and it seems to be fine, but A) I can't imagine why we'd need to support multiple items with the same value in the first place, and B) I haven't 100% wrapped my head around everything going on in this component 😅

Issue #224, opened 2019-07-02
[Combobox] Controlled value will open suggestion popover


tjallingt:
Not sure if this is mentioned anywhere but when controlling the `value` of the `ComboboxInput` the `ComboboxPopover` will start out opened.

```jsx
function App() {
  let [value, setValue] = useState("initial value");

  return (
    <Combobox>
      <ComboboxInput
        value={value}
        onChange={event => setValue(event.target.value)}
      />
      <ComboboxPopover>
        <ComboboxList>
          {results.map(result => (
            <ComboboxOption key={result} value={result} />
          ))}
        </ComboboxList>
      </ComboboxPopover>
    </Combobox>
  );
}
```

example: https://codesandbox.io/s/reachcombobox-behaviour-test-uc5qj

This seems to happen because of the transition to change triggered by these lines:
https://github.com/reach/reach-ui/blob/master/packages/combobox/src/index.js#L362-L367

Additionally controlling the onChange is not called when selecting a value using keyboard or mouse so it is necessary to also add an onSelect to the `Combobox`.

tjallingt:
Something like this might work (but would require some hacky changes to the "reducer machine")

```js
// If they are controlling the value we still need to do our transitions, so
// we have this derived state to emulate onChange of the input as we receive
// new `value`s ...[*]
if (isControlled && controlledValue !== value) {
  if (value === "") {
    // this is most likely the initial value so we want to
    // update value without transitioning to SUGGESTING
  } else {
    handleValueChange(controlledValue);
  }
}

```

marciobarrios:
I have the same problem, I fixed it having a a boolean to control when to show the popover and a `useEffect` to capture when the are suggestions:
```js
useEffect(() => {
    if (suggestions.length) {
      console.log(suggestions.length)
      setReadyToShowAutocomplete(true)
    }
  }, [suggestions])
```

tonypee:
Im also encountering this now... It seems that the component isnt designed to be controlled?

PR #227, opened 2019-07-03
[Combobox] do not suggest on initial controlled value

tjallingt:
fixes https://github.com/reach/reach-ui/issues/224

- Updated the defaultData value to `null` so we can detect the default (vs user emptied input)
- Added `CHANGE_SILENT` action which will apply the same updates as `CHANGE` without transitioning to `SUGGESTING`
  - Not sure if I need to add this to other states than `IDLE`, since I explicitly check for `value === null` I think this can only happen in `IDLE` but I might be missing something.


tjallingt:
Hiya, please let me know if the approach i took in this PR is acceptable to you. I can update the PR to no longer conflict with master.

PR #205, opened 2019-06-13
Fix menu button hover in IE 11 (fixes #201)

rebolyte:

PR #229, opened 2019-07-05
Fix to allow storybook build and view in IE11

alistairjcbrown:
When attempting to build storybook to confirm the IE 11 fix in #205, I encountered some issues.

Fixes
 - Unavailable module error when building storybook examples
 - Pulls in CommonJS build of React Spring for IE11 support


alistairjcbrown:
@ryanflorence An an aside to this PR, let me know if I can help with running through issues / PRs - we're currently using a bunch of reach-ui components internally (and so my motivation to get #205 landed), and I hope we can make use of more in the future.

PR #232, opened 2019-07-11
fixed tooltip collision with right side of screen

FIL1994:
window.innerWidth includes the width of the scrollbar.
document.documentElement.clientWidth excludes the scrollbar.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -503,7 +503,7 @@ const getStyles = (position, triggerRect, tooltipRect) => {
 const positionDefault = (triggerRect, tooltipRect) => {
   const collisions = {
     top: triggerRect.top - tooltipRect.height < 0,
-    right: window.innerWidth < triggerRect.left + tooltipRect.width,
+    right: document.documentElement.clientWidth < triggerRect.left + tooltipRect.width,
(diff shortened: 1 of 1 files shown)


chancestrickland:
@FIL1994 I'm not actually seeing any collision issues here (tested on Edge and IE11 to be sure). Would you be able to reproduce the bug and let me know what your environment looks like so I can investigate further?

FIL1994:
@chancestrickland I made an example branch here: https://github.com/FIL1994/reach-ui/tree/tooltip-position-bug-example
I tested it in Firefox and Chrome.

I changed ```window.innerWidth``` to ```document.documentElement.clientWidth``` in node_modules and it no longer collides with the scrollbar.

NateRadebaugh:
I've also seen this issue across my sites, which is blocking us from using @reach/tooltip.... we'd like to use in our menus but it can be unpredictable when the items are on the right side of the display.

NateRadebaugh:
@chancestrickland is this PR dead?

chancestrickland:
This should be addressed with https://github.com/reach/reach-ui/commit/a376daec462ccb53d33f4471306dff35383a03a5 and updated in the next release.

Issue #132, opened 2019-03-29
Dialog: onDismiss fires when selecting text and releasing mouse outside of modal


clight-fastly:
When using onDismiss, simply releasing a mouse click outside causes onDismiss to fire. Is this behavior intended? It causes issues when someone tries to select text inside.

Behavior recorded from https://ui.reach.tech/dialog#dialog-ondismiss example

![dialog-on-dismiss](https://user-images.githubusercontent.com/43175846/55256745-d4df8300-522b-11e9-9c6b-c48a9a6ede00.gif)

chancestrickland:
This should be fixed!

PR #239, opened 2019-07-24
[dialog] Fix 'onDismiss' called when mousedown and click have different targets

venikx:
In the assumption https://github.com/reach/reach-ui/issues/132 is considered unexpected, this PR should fix https://github.com/reach/reach-ui/issues/132.

### DIFF
--- a/packages/dialog/src/index.js
+++ b/packages/dialog/src/index.js
@@ -62,53 +62,55 @@ let DialogOverlay = React.forwardRef(
       ...props
     },
     forwardedRef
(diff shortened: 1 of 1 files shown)


tmcw:
I suspect that the reach-ui style would prefer a ref created via the Component² - https://ui.reach.tech/component-component#refs.

venikx:
The following PR https://github.com/reach/reach-ui/pull/246 fixes the same issue, but uses the current project coding standard, instead of hooks.

Issue #237, opened 2019-07-20
Story book doesn't start


andrew-me:
Following the readme steps:

```
git clone git@github.com:reach/reach-ui.git
cd reach-ui
yarn install
yarn bootstrap
yarn build
```
Doing a `yarn start` gives me

```

ERROR in ./packages/dialog/examples/dropdown.example.js
Module not found: Error: Can't resolve '@reach/menu-button' in '/Users/andrewedwards/Websites/reach-ui/packages/dialog/examples'
 @ ./packages/dialog/examples/dropdown.example.js 3:0-74
 @ ./packages ^((?!node_modules).)*\.example\.js$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/react/dist/server/config/polyfills.js ./node_modules/@storybook/react/dist/server/config/globals.js (webpack)-hot-middleware/client.js?reload=true ./.storybook/config.js

ERROR in ./packages/menu-button/examples/with-tooltip.example.js
Module not found: Error: Can't resolve '@reach/tooltip' in '/Users/andrewedwards/Websites/reach-ui/packages/menu-button/examples'
 @ ./packages/menu-button/examples/with-tooltip.example.js 7:0-37
 @ ./packages ^((?!node_modules).)*\.example\.js$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/react/dist/server/config/polyfills.js ./node_modules/@storybook/react/dist/server/config/globals.js (webpack)-hot-middleware/client.js?reload=true ./.storybook/config.js
```
There's a similar issue that was resolved: https://github.com/reach/reach-ui/issues/29

venikx:
The issue should be fixed in the #240 PR. I noticed it as well, when I was trying to solve another issue.

PR #240, opened 2019-07-24
Use relative paths when package deps are not directly available

venikx:
The PR fixes https://github.com/reach/reach-ui/issues/237

From what I've gathered from https://github.com/lerna/lerna/blob/master/doc/hoist.md the hoisting mechanism only makes sure there are no duplicate package installed when two lerna packages require the same packages.

So as long as no other package requires @reach/menu-button and @reach/tooltip as their depedencies these packages will never be exposed by lerna's hoisting mechanism.

### DIFF
--- a/packages/dialog/examples/dropdown.example.js
+++ b/packages/dialog/examples/dropdown.example.js
@@ -1,6 +1,6 @@
 import React from "react";
 import Component from "@reach/component-component";
-import { Menu, MenuButton, MenuList, MenuItem } from "@reach/menu-button";
+import { Menu, MenuButton, MenuList, MenuItem } from "../../menu-button";
 import "../styles.css";
 import { Dialog } from "../src/index";
 
--- a/packages/menu-button/examples/with-tooltip.example.js
+++ b/packages/menu-button/examples/with-tooltip.example.js
@@ -4,7 +4,7 @@ import React from "react";
 import "../styles.css";
 import { action } from "@storybook/addon-actions";
 import { Menu, MenuList, MenuButton, MenuItem } from "../src/index";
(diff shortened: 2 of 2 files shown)


venikx:
Adding the packages used in the storybook example as dependencies of the package is going to bloat the package (I think). Another possible solution would be to add the packages used as devDepdenencies, but I'm not sure how hoisting works then (haven't used lerna before).

mjackson:
Hi @venikx,

Thanks for the PR! I'm the new maintainer here, so please be patient with me while I get up to speed :)

IMO we definitely want to use the full package name instead of relative paths in our examples and docs. This decouples us from the file system and allows us to transform the import at build time. It also makes the repo less fragile when we move things around.

Right now it looks like we can just add `@reach/menu-button` and `@reach/tooltip` as dev deps and it should be fixed. Would you agree?

venikx:
Hi @mjackson 
Nice! Good to hear reach-ui has a new maintainer, since Ryan seemed to be busy.

That also sounds good, I was unsure what road to go on. Either devDeoendencies or relative paths. I’ll update pr in a minute.

venikx:
@mjackson Sorry for the delay. My internet has gone out since this morning. I hoped it’d be resolved in the evening so I can pull + push to update the PR :/

siddharthkp:
@venikx I ran into the same issue, so I got you: https://github.com/reach/reach-ui/pull/255

Hope you don't mind

venikx:
Nope go ahead ^^
Internet is still failing on me atm and there is no eta for when it’s going to be resolved. So thanks for working on it. This PR can be closed in favour of yours👍

Issue #236, opened 2019-07-19
Allow `null` children in MenuList


tmcw:
Right now MenuList has a strong rule that all children must be either MenuItem or MenuLink. This makes the API pretty tricky if you have any variation in your UI, like

```js
<MenuList>
    {(logic.canUnpublish(user, notebook) ||
      logic.canUnshare(user, notebook)) && (
      <MenuItem
        key="4"
        onSelect={async () => {
          await api.documentUnpublish(notebook.id);
          reload();
          closeMenu();
        }}
        className={`${buttonClass} hover-near-black`}
      >
        <UnshareIcon className="mr2" />
        {logic.notebookStatus(notebook) === "public"
          ? "Unpublish"
          : "Disable link sharing"}
      </MenuItem>)}
</MenuList>
```

In the case that this is `false`, MenuList crashes because it encounters an invalid child. The same if you have a ternary condition and have some things be `null` if they aren't relevant to that view. Ideally, MenuList is more relaxed.

Likewise, we have menus with dividers in them: `<div>` elements with borders that keep items away from each other in meaningful groups. Those dividers aren't, and probably shouldn't be, clickable, and they also shouldn't participate in keyboard shortcuts.

venikx:
Hi @tmcw 

The first part indeed seems like a bug. I'm working on it right now I can reproduce it. 
Not sure what you mean with the second part of the issue:

> Likewise, we have menus with dividers in them: `<div>` elements with borders that keep items away from each other in meaningful groups. Those dividers aren't, and probably shouldn't be, clickable, and they also shouldn't participate in keyboard shortcuts.

I tried adding a `<hr />` in between two `<MenuItem />`'s, but it seems to render what I expect it to render. Could you provide an example clarifying it further?

tmcw:
Non-MenuItem components will render, but they crashed on hover or if you try to use keyboard shortcuts to navigate the Menu, because reach-ui will [attempt to call .focus() on them](https://github.com/reach/reach-ui/blob/master/packages/menu-button/src/index.js#L49), a method that will not exist for all types.

venikx:
I’ll try the keyboard navigation in the PR I’ve created, but elements that aren’t focusable were handled already based on the things I tried in Storybook. I’ll double check the coming week or so though.

venikx:
[...]

PR #241, opened 2019-07-26
[menu-buton] ignore `null`, `false` and `undefined` <MenuList /> children

venikx:
The PR fixes https://github.com/reach/reach-ui/issues/236.

The `null` and `undefined` values are caught by the `isFocusableChildType` check. If the child of a `<MenuList />` is `false`, `React.children` converts it to a `null`, so that case is implicitly handled.

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -346,7 +346,8 @@ MenuList.propTypes = {
 
 let focusableChildrenTypes = [MenuItem, MenuLink];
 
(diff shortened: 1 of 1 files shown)


chancestrickland:
Closed in favor of https://github.com/reach/reach-ui/commit/a292285972e5797aaaa503719c1cb1d2233bd2ed. I'll add to the documentation to (hopefully) resolve any confusion here in the near future.

PR #245, opened 2019-07-30
Add note about removing CSS warning

### LINKED ISSUE #244: Ability to skip styles altogether
I love this project, and I'm using tons of its component in the apps I work for. Thanks for all of your work, efforts and dealing with all of these issues.

I typically don't use the default styles provided by the component - but I also hate the yellow console warning I get if I don't import them. 

After digging through the source code (okay, it was literally opening the CSS file, but still), I found that I can add `::root { --reach-dialog: 1 }` to avoid this warning. I still think this should be documented though!

Would you be open to me opening a PR documenting this "manual override"?

selbekk:
This commit adds a note about how you can remove the warning of a missing CSS file, if you want to implement the required styles yourself.

Fixes #244

### DIFF
--- a/www/src/pages/styling.mdx
+++ b/www/src/pages/styling.mdx
@@ -101,7 +101,6 @@ let YourMenuList = styled(MenuList)`
 ```
 
 
-
 ## CSS Selectors
 
 Because Reach UI uses regular stylesheets for its own styles you can override them like any other element. All styles use the lowest possible "specificity score", so as long as you include the component styles before your own app styles you should not run into any specificity problems.
--- a/www/src/pages/styling.mdx
+++ b/www/src/pages/styling.mdx
@@ -19,11 +19,23 @@ Styling a Reach component feels similar to styling any native element. There are
 
 # Including Base Styles
 
-Reach UI uses stylesheets for the components' base styles. You must include these styles in your app for the components to work properly
+Reach UI uses stylesheets for the components' base styles. You must include these styles in your app for the components to work properly.
 
 ## Automatically Including Styles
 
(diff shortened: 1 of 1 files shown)


www/src/pages/styling.mdx
```diff
@@ -88,6 +100,8 @@ let YourMenuList = styled(MenuList)`
 <MenuList css={absolutely}/>
 ```
 
+
```
venikx: Can you remove the unneeded whitespace?

www/src/pages/styling.mdx
```diff
@@ -19,12 +19,24 @@ Styling a Reach component feels similar to styling any native element. There are
...
+Normally, you'll receive a warning when you forget to include the matching CSS file. If you're implementing all styles yourself, you can disable this warning by setting a root CSS variable matching the package name to 1. Here's an example:
+
+```css
+:root {
+  --reach-dialog: 1;
+}
+```
+
+Note that you should tread carefully - in most cases, the styles you import are better to override than remove completely.
```
venikx: I would at this warning message before the example, making it very clear to the consumer to move forward carefully with getting rid of the base styles. I catch myself doing the following often: read docs, find example, implement based on example and ignoring the texts under the examples.

www/src/pages/styling.mdx
selbekk:
```suggestion
-
```

selbekk:
```suggestion
```

www/src/pages/styling.mdx
selbekk: Fair point!

mjackson:
Thanks for the PR @selbekk!

PR #246, opened 2019-08-01
[dialog] Only close the dialog if the click outside originated outside

tmcw:
This PR covers the same ground as #239, but uses the Component² abstraction for refs.


venikx:
I’ll close https://github.com/reach/reach-ui/pull/239 in favor of this one ;)

coreylight:
Can this be merged @ryanflorence ?

PR #249, opened 2019-08-10
fix missing dependencies in the examples

BilalBouk:
- **dialog/examples/dropdown.example.js** has a missing dependency **@reach/menu-button**

- **menu-button/examples/with-tooltip.example.js** has a missing dependency **@reach/tooltip**

### DIFF
--- a/packages/dialog/package.json
+++ b/packages/dialog/package.json
@@ -15,6 +15,7 @@
     "@reach/component-component": "^0.1.3",
     "@reach/portal": "^0.2.1",
     "@reach/utils": "^0.2.3",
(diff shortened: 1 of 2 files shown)


venikx:
These have been fixed here: https://github.com/reach/reach-ui/pull/240 and https://github.com/reach/reach-ui/pull/229 already.

Adding those to the dependencies is not a good idea, because people consuming the components will also have to to install these components (while the lib only uses uses them in storybook examples) and thus bloating the npm install unnecessarily .

BilalBouk:
closed in favor of #240 and #229

PR #254, opened 2019-08-14
add missing peer deps and use a more forgiving range

hipstersmoothie:


venikx:
Not sure if prop-types is supposed to go into peerDependencies, because when using Typescript you don’t use that package. I might be wrong though.

hipstersmoothie:
I was just following that pattern of the other components. I can them all to deps if needed.

venikx:
I agree with having consistency. So maybe a good place to start a discussion around prop-types as part of peerDependencies or not?

mjackson:
prop-types is a regular dependency because we don't want to make everyone using reach-ui install prop-types. We only use them in development for warnings.

hipstersmoothie:
@mjackson I'll open another PR with those changes, I was just following the other comps.

mjackson:
No worries, @hipstersmoothie. Those other components should use regular deps as well.

venikx:
@mjackson I was like what, when you merged, because what you said and the code was different. Then I saw the revert😅

In that case I’d make it consistent by not making it part of the peerDependencies like I mentioned earlier and use regular deps in other comps ^^

hipstersmoothie:
Updated in https://github.com/reach/reach-ui/pull/277

Issue #194, opened 2019-06-08
MenuButton: Cannot read property 'focus' of undefined


smathson:
Encountering an issue with MenuButton throwing a `Cannot read property 'focus' of undefined` error after hovering with the mouse, moving the mouse out of the Menu and then hitting the up arrow on the keyboard. The same error is not thrown when hitting the down arrow. The issue is observable in the current docs. Gif attached.

![ReachMenuFocusError](https://user-images.githubusercontent.com/932981/59141706-d8e3fb80-8966-11e9-9462-417b1a6a2c91.gif)

smathson:
My initial investigation looks like the issue is caused here:
https://github.com/reach/reach-ui/blob/master/packages/menu-button/src/index.js#L391

After the blur `state.selectionIndex` is already -1 so the `if (nextIndex !== -1)` block is called with `nextIndex = -2`.

Happy to submit a PR, I'm just not sure what the intended behavior should be. When no item is selected and the user hits the up arrow which item should be selected? The first item? I don't see any specific rules in the ARIA guidelines governing this case but:

> When a menu opens, or when a menubar receives focus, keyboard focus is placed on the first item.

...would seem like the most relevant rule.

qoalu:
Checked what the logic on Chrome dropdown is:

![Jun-12-2019 11-38-152](https://user-images.githubusercontent.com/1624045/59340610-9cb5ef80-8d06-11e9-998a-5cc41f0e40d9.gif)

Moving mouse anywhere resets keyboard control to initial state

smathson:
Cool. @ryanflorence would you accept a PR that implements the above behavior to fix the issue?

ReasonableDeveloper:
I'm having the same issue in a test using jest + react testing library.

Unfortunately, I wasn't able to reproduce this issue in a codesandbox yet. In my tests I'm simply getting the `MenuButton` by it's data attribute and clicking it via `fireEvent.click(menuButton)`. This only happens sometimes as one test that is doing the exact same thing is working fine 😕

mdgozza:
@ReasonableDeveloper I am having the exact same issue as you with react-testing-library

ReasonableDeveloper:
My failed attempt of reproducing the error in a codesandbox can be found [here](https://codesandbox.io/s/reach-ui-github-issue-cb3wf)

[...]

PR #266, opened 2019-09-04
MenuButton: fixing null focus issue on menu button

kendallgassner:
resolving:
https://github.com/reach/reach-ui/issues/194

Testing menu button using jest/react-testing-library causes a null focus error.  In the test requestAnimationFrame is called before the refs.items array has been filled with nodes. If the refs.items list is empty we should not try and focus on the null/undefined element.


venikx:
Sounds like this is fixed with this PR: https://github.com/reach/reach-ui/pull/241

kendallgassner:
@venikx unfortunately #241  does not fix my problem. In my tests I have children. It is just that the refs to the children have not finished been added to the ref.items list at the time that the first animationFrame has been called.

venikx:
@kendallgassner Thanks for checking it out and fast feedback, since I’mcurrently on holiday so I couldn’t verify it myself! 
Too bad, but I was kinda expecting it, since I didn’t really test with requestAnimationFrame. Awesome! 🎉

mjackson:
Can you please back out the changes to `yarn.lock` from this PR? Otherwise, we can go ahead and merge.

tylerkrupicka:
@mjackson can you take a look at this again?

PR #271, opened 2019-09-08
Update docs to use hooks

### LINKED ISSUE #269: Use hooks in examples instead of component^2
component^2 was really cool once it came out, but it's not the easiest way of showing how the other components work. at least not anymore.

Would you be open to a pull request refactoring the examples in the documentation to use regular hooks instead?

selbekk:
This pull request goes through all the different docs, and does the following:

- removes usage of component-component in favor of function components with hooks
- refactors class component examples to use function components and hooks
- standardize all example component names to `Example`
- Fix a few typos, broken links and what not

Fixes #269

### DIFF
--- a/www/src/pages/rect.mdx
+++ b/www/src/pages/rect.mdx
@@ -41,37 +41,33 @@ Hook that observes and returns the measurements (ClientRect) of a DOM element. P
 
 If `observe` is false, the element's rect will no longer be observed, pass it `true` again and it will. This is mostly used for things like popovers and animations, so you can usually observe when you need to, and stop observing when you don't.
 
-```jsx
-;(() => {
-  function Example() {
-    // your own ref
-    const ref = useRef()
+```.jsx
+function Example() {
+  // your own ref
+  const ref = useRef();
 
-    // pass it in to be observered
-    const rect = useRect(ref)
+  // pass it in to be observered
+  const rect = useRect(ref);
 
-    return (
-      <div>
-        <pre>{JSON.stringify(rect, null, 2)}</pre>
-        <div
-          // and then place the ref
-          ref={ref}
-          contentEditable
-          style={{
-            display: "inline-block",
-            padding: 10,
-            border: "solid 1px",
-          }}
-          dangerouslySetInnerHTML={{
-            __html: "Edit this to change the size!",
-          }}
-        />
-      </div>
-    )
-  }
-
-  return <Example />
-})()
+  return (
+    <div>
+      <pre>{JSON.stringify(rect, null, 2)}</pre>
+      <div
+        // and then place the ref
+        ref={ref}
+        contentEditable
+        style={{
+          display: "inline-block",
+          padding: 10,
+          border: "solid 1px"
+        }}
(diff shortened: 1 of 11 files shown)


selbekk:
I see I've done some unintended changes to the formatting as well with my prettier setup. Looks like there isn't a prettier config file present in the project - not sure what to do about those changes. They look the same rendered, so perhaps it's not a huge deal 😄

tomByrer:
Great idea!  Many of the examples are much shorter with hooks.

Some issues though; are the new examples comparable with Preact & Inferno?  Obviously React is the target framework, but that isn't the only JSX renderer.  Plus, some folks may want the old class-based code for their own reasons.  Possible to have both in a tabbed selector or something?  Side effect; if someone is refactoring FROM classes to fucntions+hooks, both examples could help guide them.

Thoughts anyone else?

selbekk:
These all sounds like new features to me - features worth discussing in an issue, perhaps? 🙂 I'd love to take part of that discussion :) 

Preact has support for hooks, Inferno doesn't ([yet](https://github.com/infernojs/inferno/issues/1453)).

www/src/pages/rect.mdx
```diff
@@ -42,36 +42,32 @@ Hook that observes and returns the measurements (ClientRect) of a DOM element. P
 If `observe` is false, the element's rect will no longer be observed, pass it `true` again and it will. This is mostly used for things like popovers and animations, so you can usually observe when you need to, and stop observing when you don't.
 
 ```.jsx
-(() => {
```
mjackson: Why remove the iife?

mjackson:
@selbekk Ah, we should probably add a Prettier config to this repo. Those trailing commas are going to annoy me ;)

@tomByrer If you'd like to add examples for Preact and/or Inferno let's discuss that in a separate issue. :)

selbekk:
That’s fine - sorry for budding in with my comma preferences 😄

www/src/pages/rect.mdx
selbekk: When there’s a single component in a code block, it will call it automatically. In my head, keeping the iife was just unneccesary syntax noise.

mjackson: I asked @ryanflorence about it and he said the reason he originally used the iife's was because react-live needs a component instance to be returned, not the component class itself. Is that not the case here?

selbekk: Hm okay, I don’t think that’s the case anymore at least. I’m on my phone right noe, but the gif on their readme looks like a regular function. Perhaps they invoke it for you? 🤷‍♂️ I’ll verify tomorrow morning 🙌

selbekk: Yep it's not needed anymore. You can check it yourself by going to https://ui.reach.tech/rect/ and removing the iife if you want :)

chancestrickland:
@selbekk I've since added a prettier config and updated some docs through merging older PRs. If you could help us clean up some of the formatting changes and merge conflict, I'll be happy to merge. I think this looks great!

selbekk:
@chancestrickland - done son!

PR #272, opened 2019-09-09
VisuallyHidden: Use a named function so it has a display name 

crismali:
At the moment `VisuallyHidden` shows up in the React dev tools as `Anonymous` and in snapshot tests as `Component`, but by making it a named function it'll just say `VisuallyHidden`. 👍


mjackson:
Thanks, @crismali!

PR #277, opened 2019-09-18
move prop-types to dependencies

hipstersmoothie:
@mjackson fixed 🎉 

I made sure to only add `prop-types` to packages that use it


mjackson:
Perfect, thanks!

PR #281, opened 2019-09-29
Optional scroll

hedinne:
Hi!
In some cases opening the Dialog causes the body to jump. This is caused by disabling the scroll.   

Example:

| With jump   | Without jump   |
| ------------- |-------------|
Screen capture of the screen jumping  |  Screen capture showing same action without the jump

Since this feature is mostly just for aesthetics reasons there should be away to opt out (in my opinion). 

I've added the props `removeScroll` to  `<DialogOverlay / >` and `<Dialog />`. It is set to true by default. I also added an example to Storybook and updated the docs.


hedinne:
Hey hey,
Has this PR any chance of being merged in? I don't want to keep fixing merge conflicts and testing the code if there is no chance.

hedinne:
Hi @chancestrickland and everybody ,
I'm closing this PR since there doesn't seem to be interest at this time to implement this feature. 

But I ask you to revisit my proposal. It is very weird to force people to disable scroll and add a 15px margin to the right side of the body (especially since all other styling is optional).
Having no scroll is great most of the time but sometimes it is the wrong behavior. For example, on a mobile, it is reasonable to have content that exceeds the viewport and having to add scroll within the dialog could easily be worse for the user's experience (and the design).

Sincerely, Hedinn

PR #290, opened 2019-10-07
New slider component

chancestrickland:

This pull request:

- [x] Creates a new package
- [ ] Other

If creating a new package:

  - [x] `examples` directory
  - [x] Type definitions in an `index.d.ts` file are desired but not required for the PR to be merged


packages/slider/examples/controlled-audio-progress.example.js
```diff
@@ -0,0 +1,292 @@
...
+
+  return (hours ? hours + ":" : "") + minutes + ":" + seconds;
+}
+
+function timeToMs(time) {
+  return Math.round(time * 1000);
+}
+
+// https://github.com/streamich/react-use/blob/master/src/util/createHTMLMediaHook.ts
+function useAudio(elOrProps) {
```
mjackson: Maybe put this in a separate file because it's so large?

packages/slider/index.d.ts
```diff
@@ -0,0 +1,37 @@
+declare module "@reach/slider" {
+  export type SliderProps = React.HTMLProps<HTMLDivElement> & {
```
mjackson: I think you might need a `import * as React from 'react';` so you can use `React` types? Not 100% sure.

packages/slider/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "peerDependencies": {
+    "prop-types": "^15.7.2",
+    "react": "^16.9.0",
+    "react-dom": "^16.9.0"
+  },
+  "author": "Ryan Florence <@ryanflorence>",
+  "contributors": [
+    "Chance Strickland <@chancethedev>",
+    "Michael Jackson <@mjackson>"
+  ],
```
mjackson: Let's just use `"author": "React Training <hello@reacttraining.com>"` instead of individual `author` and `contributors` fields going forward. We're like a band. We *all* write the songs!

packages/slider/styles.css
```diff
@@ -0,0 +1,113 @@
+:root {
+  --reach-slider: 1;
+  --highlight-color: #1159a6;
```
mjackson: Should we prefix all our CSS vars with `--reach-*`? I'm just thinking about people overriding this stuff in the future. Honestly not sure the best strategy to use.

packages/utils/src/index.js
```diff
@@ -36,13 +36,20 @@ if (__DEV__) {
...
 
-export let wrapEvent = (theirHandler, ourHandler) => event => {
+export const wrapEvent = (theirHandler, ourHandler) => event => {
   theirHandler && theirHandler(event);
   if (!event.defaultPrevented) {
     return ourHandler(event);
   }
 };
 
+export const callEventWithDefault = (theirHandler, ourHandler) => event => {
```
mjackson: It looks like you're not actually using this anywhere in this PR. I think we can leave it out?

packages/utils/src/index.js
```diff
@@ -36,13 +36,20 @@ if (__DEV__) {
 
 export { checkStyles };
 
-export let wrapEvent = (theirHandler, ourHandler) => event => {
+export const wrapEvent = (theirHandler, ourHandler) => event => {
```
mjackson: Just a heads up: any time you make any change to a package (no matter how small) lerna (and our new changelog generator) are going to pick up changes like this and bump the version. So, for example, this will create an entry in the changelog for this PR that bumps both `@reach/slider` *and* `@reach/utils`. Just something to think about.

www/src/pages/slider.mdx
```diff
@@ -0,0 +1,274 @@
...
+# Slider
+
+- Source: https://github.com/reach/reach-ui/tree/master/packages/slider
+- WAI ARIA: https://www.w3.org/TR/wai-aria-practices-1.1/#slider
+
+A UI input component where the user selects a value from within a given range. A Slider has a handle that can be moved along a track to change its value. When the user's mouse or focus is on the Slider's handle, the value can be incremented with keyboard controls.
+
+Sliders are composed by an outer `Slider` component with a `SliderTrack` child component. `SliderTrack` can accept `SliderTrackHighlight`, `SliderMarker` and `SliderHandle` components as children.
+
+_Pointer Events_: Slider uses pointer events for unified event handling across different pointer types (mouse, touch, stylus, etc.). As such, to override default behavior you will need to use pointer events rather than mouse or touch events. For example, if you wish to call `event.preventDefault` to override the slider handle's activation on a mouse click event, you'll want to reach for `onPointerDown` instead of `onMouseDown`. If you need to support older browsers without native pointer events, we recommend [`pepjs`](https://github.com/jquery/PEP) as a solid polyfill.
```
mjackson: We should probably say "older browsers or Safari"

packages/slider/styles.css
```diff
@@ -0,0 +1,113 @@
...
+}
+
+[data-reach-slider-marker] {
+  width: 8px;
+  height: 8px;
+  background: inherit;
+  border: 1px solid #fff;
+  border-radius: 4px;
+  transform-origin: center;
+  transition: background 200ms ease-out;
```
mjackson: This color transition seems like it's a little too opinionated for this level. It's nice, but I think it belongs in a higher level design system.

.github/CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ yarn start
 
 Next, put a file in `packages/<component-dir>/examples/<name>.example.js` and make it look like this:
 
-```jsx
+```.jsx
```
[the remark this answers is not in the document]
chancestrickland: Definitely should have been a separate PR. I'm going to work on this in the dev-website branch, but basically the distinction dictates whether a block of code renders the react-live editor.

packages/slider/styles.css
chancestrickland: Went ahead and removed this variable altogether. We aren't using CSS vars anywhere else so no need to start now.

www/src/pages/slider.mdx
chancestrickland: Added some nuance here about Safari. Pointer events are now fully supported on desktop since 13.1, and mostly supported on iOS with some minor issues that shouldn't impact our usage.

packages/slider/styles.css
chancestrickland: Removed

lerna.json
```diff
@@ -6,5 +6,6 @@
     "access": "public"
   },
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "lerna": "2.11.0"
```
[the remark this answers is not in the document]
chancestrickland: Removed

packages/slider/examples/controlled-audio-progress.example.js
chancestrickland: Moved to a new utils.js for examples.

packages/utils/src/index.js
chancestrickland: Noted. Reverted!

packages/slider/package.json
chancestrickland: Updated

packages/slider/src/index.js
```diff
@@ -0,0 +1,638 @@
...
+// https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/utils/index.js#L9
+export function assignRef(ref, value) {
+  if (ref == null) return;
+  if (typeof ref === "function") {
+    ref(value);
+  } else {
+    try {
+      ref.current = value;
+    } catch (error) {
+      throw new Error(`Cannot assign value "${value}" to ref "${ref}"`);
```
oliviertassinari: @eps1lon do you think that it would make sense to bring this variation to Material-UI?

oliviertassinari: > @eps1lon do you think that it would make sense to bring this variation of setRef to Material-UI?

oliviertassinari: At least, it's funny to see Reach using Chakra as inspiration that used Material-UI as inspiration ✌️. @eps1lon, @mbrookes it would probably be great to push https://github.com/mui-org/material-ui/issues/6218 forward.

eps1lon: I don't know what this is catching so I can't judge if we need this. Could you open an issue with an example that we don't cover properly?

oliviertassinari: @eps1lon My assumption is that it was added to improve the DX. However, it might be better handled in the prop type.
```js
const refType = PropTypes.oneOfType([PropTypes.func, PropTypes.PropTypes.object]);
```

Ok, let's wait to get a bug report/StackOverflow question that would benefit from it. Thanks for the feedback.

chancestrickland: I see it as a DX thing since a lot of folks seem to get mixed up when dealing with refs, but no first hand experience with this particular implementation yet. You'll notice that we also use Material as inspiration -- we may or may not end up leaving out the try/catch when I start dropping this hook in its place. https://github.com/reach/reach-ui/blob/1cd033a17746310ef63cdb9ae3dcb10f7e734c17/packages/utils/src/index.js#L83

Issue #193, opened 2019-06-07
Tooltips flicker when modal dialog is added


bvaughn:
Example usage are the buttons inside of this modal:
https://github.com/bvaughn/react-devtools-experimental/blob/master/src/devtools/views/Components/ComponentFiltersModal.js

![Tooltips flickering on modal display](https://user-images.githubusercontent.com/29597/59133076-c00a2480-892b-11e9-9f11-46156dc8d0af.gif)

Maybe related to #166. Let me know if you'd like me to create a smaller repro. Kind of busy at the moment but I can make the time!

FlipgridMartin:
I ran unto a similar issue, and this it's related to the `auto-id`. On first render, all of the tooltip IDs are `tooltip:null`, and then update to `tooltip:N` on the next render. Looking into why now 🙂

FlipgridMartin:
Here's a repro for the `auto-id` bug: https://codesandbox.io/s/reach-auto-id-null-repro-0w8ki

chancestrickland:
@bvaughn Prioritizing this (a little late but we'll get there). Did you happen to find a workaround for your particular case?

I do believe @FlipgridMartin is correct about auto-id. We're reviewing a few better solutions there.

FlipgridMartin:
@chancestrickland FWIW, we didn't come up with a workaround. The visible tooltips last for ~20-100ms in most instances, and since our issue is happening when a dialog was being opened, they were barely visible before they went away.

bvaughn:
> @bvaughn Prioritizing this (a little late but we'll get there). Did you happen to find a workaround for your particular case?

@chancestrickland No, but it also didn't seem high enough priority to invest much time into (assuming it would eventually be fixed upstream)

PR #293, opened 2019-10-09
useId: Refactor to prevent ID change after first render

chancestrickland:
This addresses a bug found in https://github.com/reach/reach-ui/issues/193 where useId updating from null to its initial value after the first render appears to cause some jank with tooltips. I've updated the re-pro case to test the new implementation, which renders the correct ID without an initial null value. I also spun up a quick test in Next.js to ensure consistent behavior on SSR. 

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/auto-id/examples/basic.example.js
+++ b/packages/auto-id/examples/basic.example.js
@@ -1,0 +1,20 @@
+import React from "react";
+import "@reach/dialog/styles.css";
+import { useId } from "../src/index";
+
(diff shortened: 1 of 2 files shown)


mjackson:
I think we can fix the tooltip bug another way. Spoke w @ryanflorence about it this afternoon and learned that the fact that `useId` returns `null` on the initial render is by design. I'll push another PR later that explains more.

PR #297, opened 2019-10-11
fixes #296 replace let with const where possible

### LINKED ISSUE #296: use const instead of let when variable is not reassigned
A lot of declarations are using `let` but are never reassigned. This is confusing to new contributors and makes it harder to see if a variable actually will be reassigned.

jescowuester:

This pull request:

- [ ] Creates a new package
- [x] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,5 +2,8 @@
   "extends": "react-app",
   "globals": {
     "__DEV__": "readonly"
+  },
+  "rules": {
+    "prefer-const": "error"
   }
 }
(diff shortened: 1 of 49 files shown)


mjackson:
I personally prefer to use `const` to indicate immutable _values_, not immutable _bindings_. So I'd be -1 on this change. But I'll defer to @chancestrickland.

chancestrickland:
I'm with @mjackson on this one. This is also a bit heavy-handed for a big sweeping PR, and I'd prefer to keep our lint rules a little less opinionated. Linting should keep us from making mistakes, not enforcing our opinions.

But of course, that's just, like, my opinion. 🤐

PR #298, opened 2019-10-11
Avoid extra render

krvajal:
Technically this is a side effect outside 
a `useEffect`, but it avoids the extra render
triggered by the `setState` of the current implementation

One use case in which the extra render is bad is for example when wrapping an external library
```
function Editor() {
	const id = useId()
	const editorId = `editor_${id}`
	useEffect(() => {
		const editor = new ExampleJQueryEditor(editorId)
	}) 
	
	return <div id={editorId} />
}
```

Not sure if this will cause issues in concurrent mode.

This pull request:

- [ ] Creates a new package
- [x] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/auto-id/src/index.js
+++ b/packages/auto-id/src/index.js
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useMemo } from "react";
 
 // Could use UUID but if we hit 9,007,199,254,740,991 unique components over
 // the lifetime of the app before it gets reloaded, I mean ... come on.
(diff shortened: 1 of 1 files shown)


chancestrickland:
Thanks for submitting @krvajal! We've actually got a separate branch where we're reworking auto-id completely. I'm going to hold on this for the time being as we may have that merged in the next release. If we decide to scrap that we can revisit this PR. 👍

mjackson:
The double render is intentional because ids are not supposed to be generated on the server. See [my comment here](https://github.com/reach/reach-ui/pull/319#issue-329548089) for more info.

Since others have also noticed and mentioned this, we should probably make a note about it in the docs or at least in the code so people know why we're using `useEffect` instead of `useMemo`.

chancestrickland:
We should also make sure that anywhere `useId` is used internally can be easily overridden by a prop if a user needs to bypass this behavior. I'll create a new issue for those in lieu of changes to auto-id for now.

PR #299, opened 2019-10-11
fixes #202: @reach/tooltip can be dismissed when hovered

### LINKED ISSUE #202: WCAG 2.1 Tooltip escaping
https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus

Level AA Success criterion:

> **Dismissable**
> A mechanism is available to dismiss the additional content without moving pointer hover or keyboard focus, unless the additional content communicates an input error or does not obscure or replace other content;

Does this mean that the tooltip should be dismissable on ESC key press?

jescowuester:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory


packages/tooltip/src/index.js
```diff
@@ -232,7 +232,6 @@ export function useTooltip({
   onMouseLeave,
   onFocus,
   onBlur,
-  onKeyDown,
```
chancestrickland: We need to keep this, and any event handler we use internally should be passed as the second argument in our `wrapEvent` function. This allows developers who use our components to stop our event handlers when they call `event.preventDefault`.

packages/tooltip/src/index.js
```diff
@@ -262,6 +261,25 @@ export function useTooltip({
 
   useEffect(() => checkStyles("tooltip"));
 
+  useEffect(() => {
+    const handleKeyDown = event => {
+      if (
+        event.key === "Enter" ||
+        event.key === " " ||
+        event.key === "Escape"
```
chancestrickland: I think this approach is probably OK for the escape key, but not for `Enter` or `Spacebar`. I'd probably define a separate handler inside this `useEffect` hook and attach it to document (exactly as you are doing), but leave the current handler in place (keeping it wrapped along with the user-provided `onKeyDown`.

I'll still need to test this and run it by some a11y folks since keyboard language in the WIA-ARIA spec is a bit vague here, but we'll need the above changed at minimum.

packages/tooltip/src/index.js
```diff
@@ -343,7 +351,6 @@ export function useTooltip({
     onFocus: wrapEvent(onFocus, handleFocus),
     onBlur: wrapEvent(onBlur, handleBlur),
     onMouseLeave: wrapEvent(onMouseLeave, handleMouseLeave),
-    onKeyDown: wrapEvent(onKeyDown, handleKeyDown),
```
chancestrickland: Please keep this, per the notes above

packages/tooltip/src/index.js
```diff
@@ -375,7 +382,6 @@ export default function Tooltip({
     onMouseLeave: child.props.onMouseLeave,
     onFocus: child.props.onFocus,
     onBlur: child.props.onBlur,
-    onKeyDown: child.props.onKeyDown,
```
chancestrickland: Please keep this

packages/tooltip/src/index.js
jescowuester: In that case, would it make sense to add/remove the event listener on mouseEnter/mouseLeave instead of useEffect so that keyboard users aren't affected? Does this handler also need to be preventable?

PR #317, opened 2019-10-16
Dialog: Add warnings and documentation to catch missing labels

chancestrickland:
Per the discussion in #20, I updated this branch to remove `accessibilityLabel` in favor of the web standard `aria-label`, but I kept the prop-type warnings in place and added some documentation to the Dialog page.

I would like to actually pull that documentation from the page and link to a separate new page about labeling since this applies to several components. Same thing for the prop-type validation. Considering that a phase 2 but I think this is important context for the project overall. 

This pull request:

- [ ] Creates a new package
- [ ] Other


packages/dialog/src/index.js
```diff
@@ -179,3 +201,27 @@ function useForkedRef(...refs) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, refs);
 }
+
+function validateAriaLabel(props, name, compName, ...rest) {
```
mjackson: Let's call this function `ariaLabelType` in keeping with the idea of "prop types". This function is a "type" of prop.

packages/dialog/src/index.js
```diff
@@ -179,3 +201,27 @@ function useForkedRef(...refs) {
...
+  const details =
+    "\nSee https://www.w3.org/TR/wai-aria/#aria-label for details.";
+  if (!props["aria-label"] && !props["aria-labelledby"]) {
+    return new Error(
+      "You must provide either an `aria-label` prop or an `aria-labelledby` prop to the `" +
+        compName +
+        "` component." +
```
mjackson: Instead of saying "the \`Dialog\` component" let's say "a \<Dialog> element" or just "a \<Dialog>".

packages/dialog/src/index.js
```diff
@@ -179,3 +201,27 @@ function useForkedRef(...refs) {
...
+        details
+    );
+  }
+  if (props["aria-label"] && props["aria-labelledby"]) {
+    return new Error(
+      "You provided both `aria-label` and `aria-labelledby` props to the `" +
+        compName +
+        "` component. If the a label for this component is visible on the screen, that label's component should be given a unique ID prop, and that ID should be passed as the `aria-labelledby` prop into `" +
```
mjackson: Same here. Let's say "You provided both \`aria-label\` and \`aria-labelledby\` props to a \<Dialog>"

packages/dialog/src/index.js
```diff
@@ -83,12 +86,18 @@ const DialogInner = React.forwardRef(function DialogPortal(
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
 export const DialogContent = React.forwardRef(function DialogContent(
   { onClick, onKeyDown, ...props },
   forwardedRef
 ) {
   return (
     <div
+      role="dialog"
```
mjackson: Seems weird this was missing before... isn't it kind of important?

www/src/pages/dialog.mdx
```diff
@@ -528,6 +528,46 @@ function Example(props) {
...
+    <Dialog aria-labelledby={labelId} isOpen={true}>
+      <h1 id={labelId}>Next Steps</h1>
+      <p>Follow these steps carefully!</p>
+      <button>Go Forward</button> <button>Go Back</button>
+    </Dialog>
+  )
+}
+```
+
+> NOTE: American users may see `aria-labelledby` and suspect a typo. In some regions (including the UK and Great Britain), the word is indeed spelled `labelled` instead of `labeled`. The correct HTML attribute is `aria-labelledby`.
```
mjackson: Excellent docs! Love it ❤️

packages/dialog/src/index.js
chancestrickland: I thought so too, but following the spec it should be there https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

packages/dialog/src/index.js
chancestrickland: Updated in https://github.com/reach/reach-ui/pull/317/commits/ef3e5eaa00ca3fd34bf41657c0312a6e8d0ab338

packages/dialog/src/index.js
chancestrickland: Updated https://github.com/reach/reach-ui/pull/317/commits/ae3794dcf240bd27a68688dc3f6b461953ea3d2f

PR #318, opened 2019-10-16
Simpler slider component

chancestrickland:
This PR renames the current `Slider` component to `SliderInput`, which becomes the composable API. We would then use `Slider` and a simplified version that can be used in isolation the same way an HTML range input is used. Any of the following would work

```jsx
<Slider /> 

<Slider min={0} max={200} /> 

<Slider onChange={setValue} value={value} />

<SliderInput>
  <SliderTrack>
    <SliderTrackHighlight>
    <SliderHandle />
  </SliderTrack>
</SliderInput>
```


chancestrickland:
@mjackson Friendly reminder that, if you do want to add this to a non-beta release tomorrow, we want to merge this PR first!

PR #319, opened 2019-10-18
Fix tooltips flicker

### LINKED ISSUE #193: Tooltips flicker when modal dialog is added

mjackson:
Also, added another example with tooltips rendering inside a modal to try to replicate more closely the bug described by @bvaughn in #193.

Note: It was suggested in #193 that the bug was in the auto-id package since it provides `null` ids on the initial render. According to @ryanflorence this behavior is intentional. auto-id is not supposed to provide ids on the initial render because of the non-deterministic nature of server-rendered apps. Instead, they are supposed to be applied only _after_ the initial render, to be used by screen readers.

Fixes #193
Closes #293

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/tooltip/examples/with-dialog.example.js
+++ b/packages/tooltip/examples/with-dialog.example.js
@@ -1,0 +1,48 @@
+/* eslint-disable jsx-a11y/accessible-emoji */
+import "../styles.css";
+
+import React from "react";
+import { Dialog } from "@reach/dialog";
+import Tooltip from "@reach/tooltip";
+
+export const name = "With Dialog";
+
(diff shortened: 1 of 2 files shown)


chancestrickland:
Looks good to me 👍

PR #326, opened 2019-10-22
Fix data attribute typo in alert dialog

danieltott:
Adding this in its own PR, because it _could_ be treated as a breaking change? It's a small typo on `<DialogContent>`, but if a user is using that data attribute for styling, fixing the typo could result in breaking that user's styles.

This pull request:

- [ ] Creates a new package
- [ ] Other


danieltott:
Maybe the solution is to _add_ the correct data attribute as a patch, and then in the next major release remove the attribute with the typo?

chancestrickland:
Ooof, that is...hilarious actually. We'll probably go with the patch until we release 1.0. Want to make that quick change @danieltott?

PR #329, opened 2019-10-23
New component: @reach/checkbox

chancestrickland:

This pull request:

- [x] Creates a new package
- [ ] Other

If creating a new package:

- [x] Make sure the new package directory contains each of the following, and that their structure/formatting mirrors other related examples in the project:
  - [x] `examples` directory
  - [ ] Type definitions in an `index.d.ts` file are desired but not required for the PR to be merged **TODO**


packages/checkbox/examples/group.example.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
ryanflorence: What does this illustrate other than that you can have multiple checkboxes?

packages/checkbox/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+  );
+});
+
+if (__DEV__) {
+  CustomCheckbox.propTypes = {
+    checkmarks: shape({
+      true: node,
+      false: node,
+      mixed: node
+    })
```
ryanflorence: shouldn't this be required?

packages/checkbox/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+      readOnly={readOnly}
+      type="checkbox"
+    />
+  );
+});
+
+if (__DEV__) {
+  MixedCheckbox.propTypes = {
+    checked: oneOfType([bool, string]),
+    defaultChecked: bool,
```
ryanflorence: should this be `oneOfType([bool, string])`?

packages/checkbox/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+    disabled: bool,
```
ryanflorence: Do we need the rest of these prop-types? They're all just normal checkbox proptypes (and React, afaik) doesn't give us a way to extend "checkbox proptypes".

Additionally I don't think name or value need to be required. They aren't required in normal checkboxes, what makes this different?

packages/checkbox/src/index.js
chancestrickland: I don't think so. You could style a custom checkbox without passing any additional components (with pseudo elements, for example).

ryanflorence:
I made an example of an uncontrolled custom checkbox and I'm getting this error:

```
MixedCheckbox contains an input of type checkbox with both checked and defaultChecked props. Input elements must be either controlled or uncontrolled
```

```jsx
     <CustomCheckbox
        defaultChecked={state}
        onChange={event => {
          const { checked } = event.target;
          setState(checked);
        }}
        checkmarks={{
          true: <span style={smileyStyle}>😃</span>,
          false: <span style={smileyStyle}>🙁</span>
        }}
      />
```

While we can fix this directly, I think the issue stems from a deeper composition problem that might reveal more problems down the line.

`MixedCheckbox` is entangled with `CustomCheckbox`. I think the code could be organized in a way that `CustomCheckbox` looks at its props and then decides to render a normal checkbox or a mixed checkbox, rather than mixed checkbox being aware of custom checkboxes.

`MixedCheckbox` can have a much simpler implementation because it:

1. won't have any knowledge of custom checkboxes
2. won't need to support being uncontrolled, all mixed checkboxes are necessarily controlled.

The current code has a cartesian product of requirements that must be dealt with, rather than things being encapsulated and then composed.

packages/checkbox/src/index.js
```diff
@@ -0,0 +1,266 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+export const CustomCheckbox = forwardRef(function CustomCheckbox(
+  {
+    checked: controlledChecked,
+    defaultChecked,
+    disabled: disabledProp,
+    checkmarks,
+    children,
+    id,
```
ryanflorence: any reason we're messing around with `id`, `name` here? They just get forwarded to the underlying input.

chancestrickland: The rest of the props are forwarded to the wrapper, but even for a high-level `CustomCheckbox` abstraction users will regularly need to pass `id`, `name` and `value` to the input

packages/checkbox/examples/group.example.js
chancestrickland: Not much without the `CheckboxGroup` component! I removed this one.

packages/checkbox/src/index.js
chancestrickland: Fixed

chancestrickland:
@ryanflorence I updated the MixedCheckbox to remove any reference to CustomCheckbox completely, simplifying it to its singular purpose as we discussed. Take another look when you have a moment!

ryanflorence:
If you click the readOnly checkbox twice it switches over to checked

PR #330, opened 2019-10-23
New component: @reach/accordion

chancestrickland:

This pull request:

- [x] Creates a new package
- [ ] Other

If creating a new package:

- [x] Make sure the new package directory contains each of the following, and that their structure/formatting mirrors other related examples in the project:
  - [x] `examples` directory


packages/accordion/src/index.js
```diff
@@ -0,0 +1,480 @@
...
+    return isArray(arr) ? arr : [];
+  }, [children]);
+
+  const { enabledIndexes, activeChildCount, hasActiveChild } = useMemo(() => {
+    let enabledIndexes = [];
+    let activeChildCount = 0;
+    let hasActiveChild = false;
+    for (let i = 0; i < childrenArray.length; i++) {
+      const child = childrenArray[i];
+      if (!(typeof child.type === "string" || child.props.disabled === true)) {
```
hzhu: What do you think about removing the boolean check for `child.props.disabled` since it is a boolean already?

```jsx
if (!(typeof child.type === "string" || child.props.disabled === true)) {...}

// vs.

if (!(typeof child.type === "string" || child.props.disabled)) {...}
```

packages/accordion/src/index.js
```diff
@@ -0,0 +1,480 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+export const Accordion = forwardRef(function Accordion(
+  {
+    allowMultiple: _allowMultiple = true,
+    allowToggle: _allowToggle = true,
+    as: Comp = "div",
+    children,
+    index: controlledIndex = undefined,
+    itemElement = "details",
```
mareksuscak: I wouldn't recommend defaulting to `details` as [it's breaking in Edge](https://caniuse.com/#feat=details).

packages/disclosure/styles.css
```diff
@@ -0,0 +1,51 @@
...
+[data-reach-disclosure-panel] {
+  padding: 0.5em;
+  margin: 0;
+  border: 1px solid #aaa;
+  border-top: 0;
+  border-bottom-left-radius: 0.25rem;
+  border-bottom-right-radius: 0.25rem;
+}
+
+[data-reach-disclosure-trigger] {
```
ryanflorence: I think there are too many rules in these styles, just leave it as a button and give them an example in the docs of the styles to make buttons look like divs.

ryanflorence: Actually, I think we don't need any styles at all for this one.

packages/accordion/src/index.js
```diff
@@ -0,0 +1,406 @@
...
+// Accordion
+
+export const Accordion = forwardRef(function Accordion(
+  {
+    children,
+    defaultIndex,
+    index: controlledIndex,
+    onChange,
+    readOnly = false,
+    toggle = false,
```
ryanflorence: I don't understand this prop. What does it do? Is it like, "can completely collapse"?

packages/accordion/src/index.js
```diff
@@ -0,0 +1,406 @@
...
+     * controlled state
+     */
+    setActiveIndex(controlledIndex);
+  }
+
+  return (
+    <AccordionContext.Provider value={context}>
+      <div data-reach-accordion="" ref={forwardedRef} {...props}>
+        {Children.map(children, (child, _index) =>
+          cloneElement(child, { _index })
```
ryanflorence: If we are cloning any reason to use context?

packages/accordion/src/index.js
```diff
@@ -0,0 +1,406 @@
...
+  const setButtonRef = useCallback(
+    node => {
+      ownRef.current = node;
+      if (node && Comp !== "button") {
+        setIsButtonElement(node.nodeName === "BUTTON");
+      }
+    },
+    [Comp]
+  );
+  const buttonAttributeProps = isButtonElement
```
ryanflorence: I'd say to only pass in props for a proper button. If they want to use a different kind of element they're going to need a lot more than just a couple attributes and a keyboard handler (and I don't want to get into the business of recreating all the behaviors of buttons like we do later in this code. Either they use a button, or give us a component with the same props as a button and it's your fault for screwing up the event handlers if it doesn't work like a button)

packages/accordion/src/index.js
chancestrickland: Correct. See `data-allow-toggle` in https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

packages/accordion/src/index.js
chancestrickland: We are using context values in nested components and it seemed simpler than passing those props all the way down the tree.

Issue #327, opened 2019-10-22
Why are there almost no tests?


beckerei:
## ❓Question

I was curious to contribute to `reach-ui` as we recently started to use components from it in our apps.
Checking out the code I noticed that there are almost no tests. So I don't feel very confident in contribute as I never know if I break something else.
Even more I no longer feel confident in using reach-ui components as a base in our app.

Is there a motivation behind "no tests"?

chancestrickland:
No motivation, no good reason, we just haven't. I'm planning to write them for new components as we introduce them, but if you'd like to contribute some for existing components that would be very helpful, and I'm happy to offer any guidance if you want to submit a PR.

Tests are on the todo list but @ryanflorence @mjackson and I need to coordinate on priority for that before I make any promises!

beckerei:
I'm happy to help 👍  

> I need to coordinate on priority for that before I make any promises!

chancestrickland:
@beckerei I'm good with testing-library. I don't think we'll need too much here though. Based on a quick convo with Ryan, you might find this helpful as a guide for test sequences.

- Render -> snapshot -> fire event -> snapshot
- check `document.activeElement` -> fire event -> snapshot
- check `document.activeElement`

beckerei:
Seems fine, even though I personally tend to use less snapshots as they take longer (might not matter in this case) than other assertions, and specific assertions help to understand what went wrong, while snapshots cover a lot of stuff which can change, but doesn't matter.

But I start with the proposed snapshot approach, I will check if the react test renderer is sufficient in that case and only switch to testing lib if it does not seem helpful.

chancestrickland:
I think for small reusable components snapshots can work well and shouldn't be too slow. Thanks for the assist!

beckerei:
I tried to cover all use cases of the tooltip and created a Draft PR.
There is still an issue with the `useId` which should be mocked or reset in some way before each test.

PR #331, opened 2019-10-23
[@reach/tooltip] Add tests

beckerei:
This pull request:

- [x] Other: Adding tests, in response to https://github.com/reach/reach-ui/issues/327

In this branch are two approaches how to test.

First: all snapshosts, as suggested in the referencing issue https://github.com/beckerei/reach-ui/blob/tooltip-add-tests/packages/tooltip/src/index.test.js

Second: more specific assertions https://github.com/beckerei/reach-ui/blob/tooltip-add-tests/packages/tooltip/src/index_.test.js

I personally prefer approach two, but it's not about personal test. 
It should be easy to pick up the same approach for other components too, easy to read, easy to extend.
I leave it up for you (whoever reads this) and the maintainers to decide for an approach. The other test can than be deleted

I tried to cover all us cases of the tooltip, but if anything is missing we should add these cases as well.

@alexkrolick has been extremely helpful. Thank you! 🙏


alexkrolick:
> In order to have tests that are stable the useId from @reach/auto-id should be mocked, or reset for each test

You can achieve this by using [`jest.resetModules`](https://jestjs.io/docs/en/jest-object#jestresetmodules), or by adding [`resetModules: true`](https://jestjs.io/docs/en/configuration#resetmodules-boolean) to the `jest.config.js`. I'd recommend the global config since you almost always want it. Same with `resetMocks`.

packages/tooltip/src/index.test.js
```diff
@@ -0,0 +1,86 @@
...
+jest.mock("@reach/utils", () => ({
+  ...jest.requireActual("@reach/utils"),
+  checkStyles: jest.fn()
+}));
+
+jest.useFakeTimers();
+
+describe("Tooltip", () => {
+  beforeEach(() => {
+    cleanup();
```
alexkrolick: You don't need this in the latest version of React Testing Library, it does it automatically.

beckerei: you are right 👍

beckerei:
> I'm not sure what the overall testing strategy for Reach UI is

The Snapshots are a suggestion from https://github.com/reach/reach-ui/issues/327#issuecomment-545592932 I sticked to.
But I would prefer more specific assertions as well.
The advantage of inline snapshots is that the result is visible in the test, at the same time they tend to clutter the file and make it much less readable.
With specific assertions you can ignore stuff like the positioning in tests where it doesn't matter.
No matter how we decide both of these test strategies are better than no tests.

I will incorporate your feedback later today 🙏

packages/tooltip/src/index.test.js
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import { render, cleanup, fireEvent, act } from "@testing-library/react";
+import Tooltip, { LEAVE_TIMEOUT, MOUSE_REST_TIMEOUT } from ".";
+
+jest.mock("@reach/utils", () => ({
+  ...jest.requireActual("@reach/utils"),
+  checkStyles: jest.fn()
+}));
```
ygd: 👌

beckerei:
`resetModules: true` sadly did not do the trick. I think it would still be required to reimport `"@reach/auto-id"` before each test, and because it's used by the Tooltip it would require loading the Tooltip before each test. Which I can't get to work right now.
I will start implementing a test file without snapshots instead, the auto generated ids should matter there.

alexkrolick:
@beckerei kind of weird but you can try this

Screen Shot 2019-10-25 at 12 42 53

beckerei:
@alexkrolick 
I can confirm that this resets the generated ids for each test.
However calling cleanup afterEach test does the trick:

```javascript
describe("Tooltip", () => {
  let Tooltip;
  let render, fireEvent, act, cleanup;
  beforeEach(() => {
    ({ default: Tooltip, } = require('.'));
    ({ cleanup, render, fireEvent, act } = require("@testing-library/react"));
    
    jest.useFakeTimers();
  });

  afterEach(() => {
    cleanup();
  })
)};
```

Moving `cleanup` in the `beforeEach` block does not work.

beckerei:
Tests with all Snapshots: https://github.com/beckerei/reach-ui/blob/tooltip-add-tests/packages/tooltip/src/index.test.js

Tests being more specific: https://github.com/beckerei/reach-ui/blob/tooltip-add-tests/packages/tooltip/src/index_.test.js

alexkrolick:
There must be a difference between variables declared inside the `describe` block and ones outside of it. It's unfortunate that the global setup doesn't work, because there can be subtle errors if someone forgets to add the cleanup step.

package.json
```diff
@@ -63,5 +63,8 @@
       "docs": "DOCUMENTATION",
       "enhancement": "FEATURE"
     }
+  },
+  "devDependencies": {
+    "@testing-library/jest-dom": "^4.2.0"
```
alexkrolick:
```suggestion
    "@testing-library/jest-dom": "^4.2.0",
    "@testing-library/react": "^9.3.0"
```

packages/tooltip/package.json
```diff
@@ -20,6 +20,7 @@
     "react-dom": "^16.8.0"
   },
   "devDependencies": {
+    "@testing-library/react": "^9.3.0",
```
alexkrolick:
```suggestion
```

packages/tooltip/src/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,186 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Tooltip hides on ESC 1`] = `
```
alexkrolick: One thing you can do if you end up keeping these snapshot files is add a label to them:

```js
expect(baseElement).toMatchSnapshot('after focus');
```
```
exports[`Tooltip hides on ESC: after focus 1`] =
```

beckerei:
I'm an idiot for leaving `it.only` in there and wondering why everything is okay ... 🤦‍♂

alexkrolick:
@beckerei I think this is working. https://github.com/beckerei/reach-ui/pull/2/files

Using a mock + re-requiring in beforeEach. I guess `beforeEach` executes inside `jest.isolateModules` so you don't need to explicitly add that around the require.

beckerei:
> @beckerei I think this is working. https://github.com/beckerei/reach-ui/pull/2/files

Maybe I copied it wrong from your PR but, it does not seem to work. The ids do not reset.

Furthermore I have troubles getting the two test to work (https://github.com/reach/reach-ui/pull/331/files#diff-0d85a0cf11d9802c96f656ecfc87d1dbR112) they are basically copied from the other file where they work as expected.

I had to put `jest.useFakeTimers()` inside `beforeEach` in order to work around a bug in jest https://github.com/facebook/jest/issues/5371
Adding it again to the nested describe block does not help.

alexkrolick:
>  The ids do not reset.

🤦‍♂ right

> Furthermore I have troubles getting the two test to work (https://github.com/reach/reach-ui/pull/331/files#diff-0d85a0cf11d9802c96f656ecfc87d1dbR112) they are basically copied from the other file where they work as expected.

I don't see these as working on the main branch after removing the it.only from the test above, but it does work after pasting in the cleanup/setup beforeEach/afterEach from the snapshot-based files

alexkrolick:
Ok, having thought about this a bit, it seems pretty messy, and it also seems like that other tests in both the internal Reach test suite and consumers will need a similar mechanism to reset the IDs.

An an alternative to resetting the module state, what about exposing the either the counter variable or a function wrapper for it from auto-id?

```js
import { resetCounter } from '@reach/auto-id'
import Tooltip from '.'

beforeEach(resetCounter)

test('tooltip works')
```

cc @ryanflorence

beckerei:
> An an alternative to resetting the module state, what about exposing the either the counter variable or a function wrapper for it from auto-id?

That would be an option, would add this to a global `beforeEach`.
Will check it out the next days, when I have time again to work on the failing tests.

beckerei:
A naive
```
export const resetId = () => {
  id = 0;
};
```
within a `beforeEach` does not work either, using jest with `--runInBand` option.

I just give up on the snapshot approach 😬 and fix the rest. Think it's better do discuss on a branch that has actual stable, green tests.

alexkrolick:
@beckerei this branch has passing tests for both versions https://github.com/beckerei/reach-ui/pull/3

beckerei:
> @beckerei this branch has passing tests for both versions [beckerei#3](https://github.com/beckerei/reach-ui/pull/3)

I see 👍 I'm was slightly off my notifications for this.
Thanks for all the effort ❤️

packages/tooltip/src/index.test.js
alexkrolick: Note to anyone not following the conversation: you do need it in the current setup with the re-require of the modules in beforeEach

chancestrickland:
Planning to take a look at these this week. Thanks for the assist!

chancestrickland:
@beckerei great work here, and thank you @alexkrolick for your help as well! I made a few updates and merged your commits into the testing branch. Only major change is that we opted to mock auto-id just to make things a little simpler overall. I'm going to close this PR but you'll see your work merged in after we're ready to merge testing into master.

alexkrolick:
@chancestrickland left some comments on the commit to the testing branch https://github.com/reach/reach-ui/commit/b0c5fe5cbc22e3d3b6792e18f6e9336c6b83d497#commitcomment-35861308

I don't think you want to mock autoid if you want valid a DOM structure (e.g. htmlFor => id relationships)

raunofreiberg:
What happened to these tests? I don't see them in master.

beckerei:
> What happened to these tests? I don't see them in master.

Not sure :/ 
I haven't followed along here. I'm not sure if the tests still work.

chancestrickland:
I deleted the testing branch because it fell so far behind master after the TS conversion and I've just started writing tests on master anyway. I'll get the tooltip test back up shortly.

beckerei:
Ping me if you need a helping hand.
Still happy if I can contribute.

PR #335, opened 2019-10-24
[@reach/tooltip] Simplify Tooltip statechart management

Andarist:

This pull request:

- [ ] Creates a new package
- [x] Other

Description

By definition, statecharts ignore events for which they don't have defined transitions on the active state node. So it seems to me this is counterproductive to guard against "undefined" transitions at the event dispatch site. The main reason for that is that whenever you want to tweak your statechart definition you'd have to tweak your transition call sites too

This was previously submitted as https://github.com/reach/reach-ui/pull/176 but that PR got closed because of housekeeping purposes. I still believe this change is valid and beneficial, so I'm resubmitting this.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -144,11 +144,10 @@ function transition(action, newContext) {
 
   // Really useful for debugging
   // console.log({ action, state, nextState, contextId: context.id });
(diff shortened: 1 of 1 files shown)


Andarist:
@chancestrickland friendly 🏓

PR #340, opened 2019-10-28
docs: Add popover to list of components

brunoxd13:
Add popover to list of components on readme file.

This pull request:

- [ ] Creates a new package
- [X] Other

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ The components to be built come from the the [Aria Practices Design Patterns and
 | 🛠      | Listbox              |
 | 🛠      | Menu or Menu bar     |
 | ✅     | Menu Button          |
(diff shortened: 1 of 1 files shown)


chancestrickland:
I don't believe popover is intended to be a part of our public API, more of an internal tool. I'll confirm that this was Ryan's intent, but we'll likely keep it unlisted if that is the case.

ryanflorence:
Correct. We'll have a public popover soon though. If we made this public right now there would be accessibility issues around pretty much every usage of it.

There's a spectrum of "popovers"

`no interaction: tooltip < --- menu | restricted interaction | popup --- > dialog: all interaction`

We want to get it right.

PR #342, opened 2019-10-28
New component API notes 📝

chancestrickland:


api-notes.md
```diff
@@ -0,0 +1,325 @@
...
+  slidesShown?: number; // default: 1
+};
+```
+
+## 🛠: Disclosure
+
+> NOTE: This should follow the Accordion, as it's essentially the same basic behavior with less flexibility and no custom focus management.
+
+```jsx
+<Disclosure>
```
ryanflorence: Not sure why we'd need a parent element here? Don't think we need to necessarily group them. I'd imagine we only need:

```
<Disclosure>
  <DisclosureButton>Open the box</DisclosureButton>
  <DisclosureContent>Some content</DisclosureContent>
</DisclosureItem>
```

api-notes.md
```diff
@@ -0,0 +1,325 @@
...
+  <DisclosureItem>
+    <DisclosureTrigger>Open the box</DisclosureTrigger>
```
ryanflorence: Would probably go with `<DisclosureButton/>` and render a `button` so we don't have to screw around with keyboard/focus/click etc.

api-notes.md
```diff
@@ -0,0 +1,325 @@
...
+TBD
+
+## ❓: Link
+
+TBD
+
+## 🛠: Listbox
+
+```jsx
+<Listbox>
```
ryanflorence: Not sure what this wrapper `Listbox` component is for? Much like the disclosure one. Listboxes aren't related to eachother any more than `<select>`.

api-notes.md
```diff
@@ -0,0 +1,325 @@
...
+  <ListboxArea>
```
ryanflorence: Same here, not sure what this would do

PR #344, opened 2019-10-29
Make visually hidden render as span

qoalu:
Refer to SO question: https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button/12982334

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/visually-hidden/src/index.js
+++ b/packages/visually-hidden/src/index.js
@@ -12,7 +12,7 @@ let style = {
 };
 
 function VisuallyHidden(props) {
-  return <div style={style} {...props} />;
+  return <span style={style} {...props} />;
 }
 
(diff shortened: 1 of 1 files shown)


chancestrickland:
There are use cases outside of labeling buttons/links for visually hidden text, but point received. We should probably use an `as` prop to allow an easy override here either way.

chancestrickland:
@qoalu BTW, you'll need to add `display: 'block'` to the style object as well, since the properties used to hide the element won't have an effect on inline elements.

qoalu:
thanks for the quick turnaround! 

An "as" prop could be helpful, however this solution with span and display block should suffice for all use-cases.

PR #361, opened 2019-11-08
[@reach/tabs] Decouple UI Hierarchy with Context

indiesquidge:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

## Tabs: Now with Context

`<Tabs>` and its constituents have now been built using [Context] instead of `cloneElement`, which relieves the direct parent-child restriction of the current implementation for `<Tabs>/<TabList>` and `<Tabs>/<TabPanels>` *. It also now exposes [styling state] with a `useTabState` [hook].

### Motivation

> NOTE: this is influenced from the _Implicit State with Compound Components and Context_ [lecture] video, the [Reach UI Philosophy Guide], and my own experience working with `<Tabs>`.

The current `<Tabs>` is great, but suffers a limitation in which all of its components must have a direct parent-child relationship. But what if you want to insert other components in between?

```jsx
<Tabs>
  <div className="this-breaks-tabs">
    <TabList>
      <Tab />
    </TabList>
  </div>

  <div className="also-breaks-tabs">
    <TabPanels>
      <TabPanel />
    </TabPanels>
  </div>
</Tabs>
```
By adding a couple seemingly innocuous divs we lose the active tab indicator and click handler functionality.

Or what if we want to add sibling elements to the tab list or tab panels?

```jsx
<Tabs>
  <TabList>
    <Tab />
  </TabList>

  {/* The div below would be passed unnecessary props from Tabs when cloned … whoops! */}
  <div className="foo">Some sibling content to tab list and tab panels</div>

  <TabPanels>
    <TabPanel />
  </TabPanels>
</Tabs>
```

We start passing [Tabs-specific props] to native (or composite) components that shouldn't have those additional props!

For both situations, using context allows us to decouple that UI hierarchy.

---

\* It's worth noting that this PR doesn't remove the necessity for the direct parent-child relationship between `<TabList>` and each `<Tab>` as well as `<TabPanels>` and each `<TabPanel>`. In order to know which tab is active, we have to know about the order of each `<Tab>` and `<TabPanel>` (currently using its index as a direct child). In order to decouple this relationship it would likely require significantly more overhaul with a kind of context registering. One solution could be to have each `<Tab>` and `<TabPanel>` call some register function (passed down to them through TabContext) when they mount, and the context provider would return the index to the child.

UPDATE: I've created a separate commit that removes the necessity for the direct parent-child relationship between `<TabList>` and each `<Tab>` as well as `<TabPanels>` and each `<TabPanel>`. You can see what these changes would look like [here](https://github.com/indiesquidge/reach-ui/compare/tabs-with-context...indiesquidge:tabs-with-context-part-2?expand=1). (Please read the commit message in the PR description to note the caveats this approach would have when React running in [Concurrent Mode].)

[Context]: https://reactjs.org/docs/context.html
[styling state]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#styling-states-with-pseudo-pseudo-selectors
[hook]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#exposing-state-for-js-styling
[lecture]: https://courses.reacttraining.com/courses/250055/lectures/3897321
[Reach UI Philosophy Guide]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7
[Concurrent Mode]: https://reactjs.org/docs/concurrent-mode-intro.html
[Tabs-specific props]: https://github.com/reach/reach-ui/blob/c20cdf7f2dd6b9eabe5461138daf320e2833aa41/packages/tabs/src/index.js#L45-L64

### DIFF
--- a/packages/tabs/examples/with-non-tabs-components-interweaved.example.js
+++ b/packages/tabs/examples/with-non-tabs-components-interweaved.example.js
@@ -1,0 +1,52 @@
+import "@reach/tabs/styles.css";
+
+import React from "react";
+import { Tabs, TabList, Tab, TabPanels, TabPanel } from "@reach/tabs";
+
+export const name = "With Non-Tabs Components Interweaved";
+
+const tabsStyle = {
+  width: 400,
+  boxShadow: "1px 1px 5px hsla(0, 0%, 0%, 0.25)"
+};
+
+const tabListWrapperStyle = {
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "space-between",
+  border: "1px solid darkslategray",
+  padding: "4px 0"
+};
+
+export const Example = () => (
+  <Tabs style={tabsStyle}>
+    <div style={tabListWrapperStyle}>
+      <TabList style={{ margin: "0 16px" }}>
+        <Tab>One</Tab>
+        <Tab>Two</Tab>
+        <Tab>Three</Tab>
+      </TabList>
+      <div style={{ margin: "0 16px", textAlign: "right" }}>
+        Here is content styled alongside the tab list
+      </div>
+    </div>
+
+    <div style={{ background: "ghostwhite", padding: "16px" }}>
+      <div style={{ textAlign: "center" }}>
+        Here is content above tab panels but styled with it.
+      </div>
+      <TabPanels>
+        <TabPanel>
+          <h1>one!</h1>
+          <button>yo</button>
+        </TabPanel>
+        <TabPanel>
+          <h1>two!</h1>
+        </TabPanel>
+        <TabPanel>
+          <h1>three!</h1>
+        </TabPanel>
+      </TabPanels>
+    </div>
+  </Tabs>
+);
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -75,29 +75,32 @@ export const Tabs = forwardRef(function Tabs(
 
   const [selectedIndex, setSelectedIndex] = useState(defaultIndex || 0);
 
+  const tabsContext = React.useMemo(
+    () => ({
+      selectedIndex: isControlled ? controlledIndex : selectedIndex,
+      _id: id,
+      _userInteractedRef,
+      _selectedPanelRef,
+      _onFocusPanel: () =>
+        _selectedPanelRef.current && _selectedPanelRef.current.focus(),
+      _onSelectTab: readOnly
+        ? () => {}
+        : index => {
+            _userInteractedRef.current = true;
+            onChange && onChange(index);
+            if (!isControlled) {
+              setSelectedIndex(index);
+            }
+          }
+    }),
+    [id, controlledIndex, isControlled, onChange, readOnly, selectedIndex]
+  );
+
   useEffect(() => checkStyles("tabs"), []);
 
   return (
     <Comp data-reach-tabs="" ref={ref} {...props}>
-      <TabsContext.Provider
-        value={{
-          selectedIndex: isControlled ? controlledIndex : selectedIndex,
-          _id: id,
-          _userInteractedRef,
-          _selectedPanelRef,
-          _onFocusPanel: () =>
-            _selectedPanelRef.current && _selectedPanelRef.current.focus(),
-          _onSelectTab: readOnly
-            ? () => {}
-            : index => {
-                _userInteractedRef.current = true;
-                onChange && onChange(index);
-                if (!isControlled) {
-                  setSelectedIndex(index);
-                }
-              }
-        }}
-      >
+      <TabsContext.Provider value={tabsContext}>
         {children}
       </TabsContext.Provider>
     </Comp>
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -28,6 +28,15 @@ const TabsContext = createNamedContext("Tabs");
 const TabContext = createNamedContext("Tab");
 const TabPanelContext = createNamedContext("TabPanel");
 
+export const useTabState = () => {
+  const index = useContext(TabContext);
+  const { selectedIndex } = useContext(TabsContext);
+
+  return {
+    isSelected: index === selectedIndex
+  };
+};
+
 ////////////////////////////////////////////////////////////////////////////////
 // Tabs
 
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -1,6 +1,7 @@
 import React, {
-  cloneElement,
+  createContext,
   forwardRef,
+  useContext,
   useEffect,
   useRef,
   useState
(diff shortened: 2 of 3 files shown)


packages/tabs/src/index.js
```diff
@@ -1,9 +1,35 @@
...
 import warning from "warning";
 import { wrapEvent, useUpdateEffect, makeId, useForkedRef } from "@reach/utils";
 import { useId } from "@reach/auto-id";
 
+const createNamedContext = name => {
+  const context = createContext();
+  context.displayName = name;
+
+  return context;
+};
```
indiesquidge: Not sure if this `createNamedContext` is relevant for this repo. It was copied over from [React Router](https://github.com/ReactTraining/react-router/pull/6479/files#r233653953) based on the comment left at implementation:

> this avoids static property assignments by hiding it in a tree-shakeable function call (this was preventing tree shaking for simialr reason as .defaultProps)

packages/tabs/src/index.js
```diff
@@ -192,12 +218,12 @@ export const Tab = forwardRef(function Tab(
       data-reach-tab=""
       ref={ref}
       role="tab"
-      id={makeId("tab", _id)}
+      id={makeId("tab", makeId(_id, index))}
       tabIndex={isSelected ? 0 : -1}
       aria-selected={isSelected}
-      aria-controls={makeId("panel", _id)}
+      aria-controls={makeId("panel", makeId(_id, index))}
```
indiesquidge: The composed call to `makeId` is used here because we have [removed](https://github.com/reach/reach-ui/pull/361/files#diff-6c8cbe1adb417474a2ffacc00dbd2c22L107) it happening from `<TabList>` and `<TabPanels>` when it was cloned to their children. Happy to move it to a variable or something to make it easier to look at if desired.

packages/tabs/examples/with-wrappers.example.js
```diff
@@ -0,0 +1,37 @@
...
+        <TabPanel>
+          <h1>two!</h1>
+        </TabPanel>
+        <TabPanel>
+          <h1>three!</h1>
+        </TabPanel>
+      </TabPanels>
+    </div>
+  </Tabs>
+);
```
indiesquidge: I'm not fully convinced that this example is the best display of showing the decoupling this PR affords. I'm happy to take suggestions on better examples if people have them.

Here's a gif of the example above:

tabs

packages/tabs/src/index.js
```diff
@@ -1,9 +1,35 @@
...
+const TabPanelContext = createNamedContext("TabPanel");
+
+export const useTabState = () => {
+  const index = useContext(TabContext);
+  const { selectedIndex } = useContext(TabsContext);
+
+  return {
+    isSelected: index === selectedIndex
+  };
+};
```
indiesquidge: Don't know if Reach expects clients to be on React 16.8+, but we could add in a development warning for it the way [React Router](https://github.com/ReactTraining/react-router/blob/ea44618e68f6a112e48404b2ea0da3e207daf4f0/packages/react-router/modules/hooks.js#L10-L15) does if we'd like

```js
export const useTabState = () => {
  if (__DEV__) {
    invariant(
      typeof useContext === "function",
      "You must use React >= 16.8 in order to use useTabState()"
    );
  }

  const index = useContext(TabContext);
  const { selectedIndex } = useContext(TabsContext);

  return {
    isSelected: index === selectedIndex
  };
};
```

indiesquidge:
Looks like the current `master` branch is [failing in CI](https://travis-ci.com/reach/reach-ui).

cc: @chancestrickland.

packages/tabs/src/index.js
indiesquidge: Definitely a brain fart moment here … obviously we require 16.8 for this repo, most of the components are using hooks.

packages/tabs/examples/with-non-tabs-components-interweaved.example.js
```diff
@@ -0,0 +1,52 @@
...
+
+const tabListWrapperStyle = {
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "space-between",
+  border: "1px solid darkslategray",
+  padding: "4px 0"
+};
+
+export const Example = () => (
```
indiesquidge: Not sure if this example is the best display of showing the decoupling this PR affords. I'm happy to take suggestions on better examples if people have them.

Here's a gif of the example above:

tabs-with-context

indiesquidge:
@chancestrickland I'd love a review of this when you have some time. I see work has been done on Tabs a few times in the past month, and I'd like to help contribute to this great library. I think the additions I've made in this PR are helpful (e.g. #406 wouldn't have needed to be [patched]) and only add functionality to the package—the changes are fully backwards compatible and do not change the Tabs API.

Please let me know how I can help move this forward, thank you ☺️

[patched]: https://github.com/reach/reach-ui/commit/0ce819b06db37f0d00fde1b8f0e38c05fe582419

chancestrickland:
Hey @indiesquidge, sorry for the delay here! I agree with the motivation here and understand the benefits. The main blocker right now is that we are having a larger discussion about how we handle a lot of these components with nested components, as each approach has tradeoffs and we haven't quite settled on the right answer yet. Happy to review your work this week, definitely appreciate the contribution!

indiesquidge:
Hi @chancestrickland. Thanks for the reply! I know how busy y'all can get.

That's awesome to hear that a discussion is going on around best practices for nested components that need to communicate with each other. Over the holidays I went through Ryan's posts on [descendants] and [ListBox], both of which relate to this broader conversation and offered up ideas around the trade-offs for these more complex components. If there are more public forums where y'all are discussing this, I'd love to be part of that conversation!

That said, I believe only my changes in [part 2] are affected by these trade-offs (see the commit description hydrated in the PR textarea for my thoughts around this). The changes in this PR luckily do not deal with descendants needing to know about things like their index, only with adding in the ability to have elements as siblings of `<TabList>` and `<TabPanels>`.

Happy to hear your thoughts when you have a chance to review 🙂

[descendants]: https://gist.github.com/ryanflorence/10e9387f633f9d2e6f444a9bddaabf6e
[ListBox]: https://gist.github.com/ryanflorence/f7d435084313f5d40095ffc0d5c0ad9d
[part 2]: https://github.com/indiesquidge/reach-ui/compare/tabs-with-context...indiesquidge:tabs-with-context-part-2?expand=1

chancestrickland:
@indiesquidge Just FYI, we've introduced a new descendants hook we're trying out and I got Tabs updated to use it in #140ee1a. Took a few good ideas from your PR in the process. We will likely make some adjustments to the implementation details but the behavior in the next release should take care of these concerns. I included your example as well. Thanks again for the contribution, going to close this for now!

Issue #321, opened 2019-10-18
Fix issues with how IDs are handled


chancestrickland:
### Current Behavior

There are a few places in the project where IDs are auto-generated on the client. This may be undesirable for a few reasons (users who disable JS lose important accessibility features, potential race condition bugs when used with non-React libs, two examples).

Further, in our auto-generated IDs we often use colons as delimiters. `document.querySelector` throws an error when receiving a string with a colon.

```
document.querySelector('#id:1')
SyntaxError: '#id:1' is not a valid selector
```

### Expected behavior

We should 1) allow users to set IDs via props everywhere we are generating them currently, and 2) we should remove colons from our generated IDs.

danieltott:
@chancestrickland The first part makes sense, but can you point out where there are colons in generated IDs? I have found three components that use generated IDs: alert-dialog, menu-button, and skip-nav (although with skip-nav it's not generated, it's hard-coded). But it seems like when creating an ID, they're being concatenated with `-`, not `:` (``labelId: `alert-dialog-${genId()}`,`` for example. I'm sure I'm missing something because I'm still getting my bearings in the source code.

chancestrickland:
@danieltott Sounds good -- I created a new branch for this but that's as far as I've gotten. Happy to take a look when you have a PR ready.

danieltott:
@chancestrickland For updating components to allow for ID overriding, would it make sense to create a series of PRs per Reach component? Or would you prefer a single, larger PR?

chancestrickland:
@danieltott Single PR is fine. We typically release all at once and only the updated packages will get a version bump.

chancestrickland:
@danieltott We try to handle attributes like `aria-controls` automatically, so even if the user passes their own ID to the input they shouldn't need to worry about its controller component. In this particular case we still need to allow the user to pass an ID for combobox (working on that one), but I don't see any value in allowing them to override `aria-controls`.

danieltott:
@chancestrickland Sounds good. I should have a PR up early this week.

PR #368, opened 2019-11-13
Fix issues with how IDs are handled

danieltott:

This pull request:

- [ ] Creates a new package
- [ ] Other

This started as a fix for #321

I debated creating this PR; it steps on the heels of #365 (basically overriding everything that that PR does). Ultimately I think it's worth it though, it handles some cases that #365 doesn't and allows some more elements to have custom IDs

In general, the idea is for elements where we have to know the ID (for `aria-*` attributes usually), we manage that ID in the parent and save it in state/context, then allow for it to be overridden in the child component.

@chancestrickland If you'd rather close this and concentrate on #365 that is fine; I just wanted to get it up there. I could offer some feedback based on this work over there if you would prefer that route.

### DIFF
--- a/packages/alert-dialog/examples/id-prop.example.js
+++ b/packages/alert-dialog/examples/id-prop.example.js
@@ -1,0 +1,53 @@
+import React, { useRef, useState } from "react";
+import {
+  AlertDialogOverlay,
+  AlertDialogContent,
+  AlertDialogLabel,
+  AlertDialogDescription
+} from "@reach/alert-dialog";
+
+export let name = "ID Props";
+
+export function Example() {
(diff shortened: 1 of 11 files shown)


chancestrickland:
@danieltott Thanks for submitting, and sorry we sort of crossed paths here! After chatting with Ryan about our approach we landed on limiting the scope a bit from allowing children to override IDs. While I do still think it's a more ideal API, having to sync and update the values through context could be error prone and ultimately not worth it, considering what you're most likely after is deterministic IDs, not necessarily custom IDs.

I'll close this, but if you still think we can make some improvements based on what we merged in #365 I'm happy to review if you want to open a new issue with your feedback.

PR #377, opened 2019-11-15
tooltip: add document event listener to hide tooltip on escape

chancestrickland:
This PR addresses https://github.com/reach/reach-ui/issues/202. We move the event listener to the document so escaping at any point will dismiss the tooltip.

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -270,13 +270,11 @@ export function useTooltip({
 
   useEffect(() => {
     const listener = event => {
-      if (event.key === "Escape" || event.key === "Esc") {
-        console.log("whoa");
-        switch (state) {
-          case VISIBLE: {
-            transition("selectWithKeyboard");
-          }
-        }
+      if (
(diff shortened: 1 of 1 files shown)


packages/tooltip/src/index.js
```diff
@@ -268,6 +268,21 @@ export function useTooltip({
 
   useEffect(() => checkStyles("tooltip"));
```
ryanflorence: Could simplify this a bit:

```
useEffect(() => {
    const listener = event => {
      if ((event.key === "Escape" || event.key === "Esc") && state === VISIBLE) {
        transition("selectWithKeyboard")
      }
    };
    document.addEventListener("keydown", listener);
    return () => document.removeEventListener("keydown", listener);
  }, []);
```

chancestrickland: Yeah but then my very important, totally helpful `console.log("whoa")` wouldn't work!

Issue #199, opened 2019-06-09
[ComboBox] optionsRef approach unstable


Andarist:
Pushing to the `optionsRef` in `ComboboxOption` is unstable - if we trigger a rerender on the Option without triggering `ComboboxList` rerender (which clears `optionsRef`) we might just push and push onto it making any logic based on that array unstable.

amannn:
Same is true when `ComboboxList` renders, but `ComboboxOption` doesn't due to `React.memo`.

I guess this is related: https://github.com/reach/reach-ui/issues/354

chancestrickland:
Can either of you share a CodeSandbox link the demos the problems with our approach?

Andarist:
@chancestrickland I couldn't reproduce it **right now** because of this https://github.com/reach/reach-ui/pull/379 , but once you fix that it will be possible to reproduce it with code like this:
https://codesandbox.io/s/loving-margulis-lx5zs

PR #379, opened 2019-11-17
[combobox] Fix issue with Combobox context not being memoized correctly

Andarist:
This pull request:

When trying to reproduce a problem described in https://github.com/reach/reach-ui/issues/199 I've noticed that it currently cannot be reproduced because:
- [`transition`](https://github.com/reach/reach-ui/blob/f22cdf931d52384205575f3060933241056e4d1e/packages/combobox/src/index.js#L813) is an inline function, so it has a fresh reference for each render
- but it's used as input for `useMemo` [here](https://github.com/reach/reach-ui/blob/f22cdf931d52384205575f3060933241056e4d1e/packages/combobox/src/index.js#L282), which makes `useMemo` yield a new reference each render and thus constantly updating the context value

I've fixed it by creating a "proxy" callback to this inline one. This way that "proxy" can have a stable reference across renders but still call "latest" version of the inline function thanks to using a ref under the hood.

I believe this solution is better than using `useCallback` directly because I think that `transition`'s reference shouldn't ever update, it should be considered stable - similarly how we can assume `setState` & `dispatch` to be stable.

### DIFF
--- a/packages/utils/src/index.js
+++ b/packages/utils/src/index.js
@@ -1,4 +1,5 @@
-import { useRef, useMemo, useEffect } from "react";
+import { useRef, useMemo, useEffect, useCallback } from "react";
+import useLatest from "use-latest";
 
 let checkedPkgs = {};
 
@@ -94,6 +95,11 @@ export function useUpdateEffect(effect, deps) {
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, deps);
+}
+
+export function useStableCallback(callback) {
+  const latestRef = useLatest(callback);
+  return useCallback((...args) => latestRef.current(...args), [latestRef]);
 }
 
 export function wrapEvent(theirHandler, ourHandler) {
--- a/packages/combobox/src/index.js
+++ b/packages/combobox/src/index.js
@@ -21,7 +21,12 @@ import React, {
   useState
 } from "react";
 import PropTypes from "prop-types";
-import { makeId, wrapEvent, useForkedRef } from "@reach/utils";
+import {
+  makeId,
+  wrapEvent,
+  useForkedRef,
+  useStableCallback
+} from "@reach/utils";
 import { findAll } from "highlight-words-core";
 import escapeRegexp from "escape-regexp";
 import { useId } from "@reach/auto-id";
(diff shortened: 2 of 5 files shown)


packages/utils/src/index.js
```diff
@@ -1,4 +1,5 @@
-import { useRef, useMemo, useEffect } from "react";
+import { useRef, useMemo, useEffect, useCallback } from "react";
+import useLatest from "use-latest";
```
Andarist: If you want - I can inline this dependency. It's really simple but my personal opinion is that it's good to share common code (even if it's short like this one), but I know many people prefer inlining for cases like this. Disclaimer though - it's written by me 😉 I actually also probably will extract `useStableCallback` to a separate library as well, just didn't have time to do this right now.

chancestrickland:
I'm still unsure what bug this is actually addressing. So you're saying that, if we merge this PR, then we'll have a bug in combobox per #199? But we don't currently have any bugs in combobox described in #199, is that correct? What problem does this PR ultimately solve for us today?

Andarist:
> What problem does this PR ultimately solve for us today?

I assume this useMemo:
https://github.com/reach/reach-ui/blob/f22cdf931d52384205575f3060933241056e4d1e/packages/combobox/src/index.js#L265
has been introduced for a reason. Right now you can either fix its dependencies (by merging this PR or some alternative way of fixing this `useMemo` dependencies) or remove this `useMemo` entirely.

> So you're saying that, if we merge this PR, then we'll have a bug in combobox per #199? But we don't currently have any bugs in combobox described in #199, is that correct?

I've found out about this issue here when I've tried to repro #199 in a very similar way as I've been reproducing it in the past - when `useMemo` was memoized correctly. When you force whole combobox-related stuff to rerender on each change then the reported issue goes away as it was specifically about memoized components in the middle breaking things. Question is - do you want to rerender everything in unrelated changes?

chancestrickland:
@Andarist Understood, thanks for your explanation and the PR!

I think for now, to keep it simple, we are just going to remove `useMemo` altogether. The biggest reason is because we will be rolling out a rewrite of combobox in the coming weeks, and adding complexity + a new dependency here is less desirable ATM. We can circle back to this approach if we need it for the optimization once the next version is released!

Andarist:
@chancestrickland cool, good to see that useMemo removed for now 👍 When you get to refactoring this component, I would appreciate giving me a ping - would love to see how you are going to do it 😃

Issue #380, opened 2019-11-18
[MenuList] Additional key handling for tab, home, end


jylertones:
### Current Behavior

Currently, when a `MenuButton` is activated, a `MenuList` appears. The key event handling for `ArrowUp`, `ArrowDown` and `Escape` are correct as done in MenuItem: https://github.com/reach/reach-ui/blob/master/packages/menu-button/src/index.js#L412.

However, when the `Home`, `End`, or the `Tab` key are pressed, nothing happens

### Expected behavior

Under the WAI-ARIA Authoring Practices, several other keystrokes also have expected behaviors:
https://www.w3.org/TR/wai-aria-practices-1.1/#menu

> Home: If arrow key wrapping is not supported, moves focus to the first item in the current `menu` or `menubar`.
> End: If arrow key wrapping is not supported, moves focus to the last item in the current `menu` or `menubar`.
> Tab: Moves focus to the next element in the tab sequence, and if the item that had focus is not in a `menubar`, closes its `menu` and all open parent `menu` containers.

### Reproducible example

https://ui.reach.tech/menu-button#menu

### Suggested solution(s)

It does not look like the Up/Down arrow keys cycle to the first/last when they are at the beginning/end of the list, so Home/End should be straightforward.

For tab, I'm guessing that there is an intentional reason why it is ignoring those key presses. I would expect that one would probably just remove the current behavior for the `MenuItem` - remove the `event.preventDefault()` that happens. This should have the effect that pressing tab would activate the next `MenuItem`, or if it it on the last item, it would focus the next item on the page and the menu would close? (In my mind, the spec is less clear here - and hopefully there is no expectation that `Shift+Tab` would reopen the menu to the last item, as I've never seen any product behave that way.)

### Additional context

If we agree on my suggested solution, I'm happy to submit a PR for this

### Your environment

| Software                 | Name/Version(s) |
| ------------------------ | --------------- |
| **[REACH_PACKAGE_NAME]** | MenuButton | 
| React                    | 16.12.0 |
| Browser                  | Firefox 71, Chrome 78 |
| Assistive tech           | Keyboard | 
| Node                     | N/A
| npm/yarn                 | N/A
| Operating System         | Linux, Win10 |

chancestrickland:
[...]

PR #381, opened 2019-11-19
on home, tab, end button press functionality added for menubutton

prasad5795:
#380 issue resolved

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -425,6 +425,18 @@ export const MenuItems = forwardRef(function MenuItems(
           }
         } else if (event.key === "Tab") {
           event.preventDefault(); // prevent leaving
+          const nextIndex = state.selectionIndex + 1;
+          if (nextIndex !== focusableChildren.length) {
+            setState({ selectionIndex: nextIndex });
+          } else {
+            setState({ selectionIndex: 0 });
+          }
+        } else if (event.key === "Home") {
+          event.preventDefault(); // prevent window scroll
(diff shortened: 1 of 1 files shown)


packages/menu-button/src/index.js
```diff
@@ -425,6 +425,18 @@ export const MenuItems = forwardRef(function MenuItems(
           }
         } else if (event.key === "Tab") {
           event.preventDefault(); // prevent leaving
+          const nextIndex = state.selectionIndex + 1;
```
chancestrickland: > Tab: Moves focus to the next element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.

The next menu item in the list is not the same as the next element in the tab sequence. We don't assume that the item is in a menubar, so ultimately the developer using `MenuButton` would need to add tab behavior if that were the case (at least until we build out a `MenuBar` component one of these days)

chancestrickland: For standalone menu buttons I believe tab should just close the menu and focus whatever element would come after the menu button in normal tab order.

prasad5795: so can you please explain me what do you mean by next element in tab sequence , because originally on press of tab it was working as per your explanation i.e. "For standalone menu buttons I believe tab should just close the menu and focus whatever element would come after the menu button in normal tab order." so if that's the required behavior then I will remove that code

chancestrickland:
This functionality was added in 0.7.0, though to note we did not change the tab behavior. We're still discussing this internally before we implement a change there.

Issue #357, opened 2019-11-06
ComboboxList does not scroll along when navigating with keyboard


raunofreiberg:
### Current Behavior

![image](https://i.gyazo.com/054c4f6807e73a9db84b977ed8941a05.gif)

### Expected behavior

The Combobox should scroll along with the focused item.

### Reproducible example

https://codesandbox.io/s/upbeat-feynman-b4gge

### Your environment

| Software                 | Name/Version(s) |
| ------------------------ | --------------- |
| **[REACH_PACKAGE_NAME]** |
| React                    |
| Browser                  |
| Assistive tech           |
| Node                     |
| npm/yarn                 |
| Operating System         |

chancestrickland:
Starting a discussion here, but I think this is probably best left to app code. If you need a fixed height popover, you should also be able to measure DOM nodes and handle scrolling with our API as needed. I'm not sure a fixed popover is a great UI pattern here for a few reasons, and I'd hate for us to bake in an opinionated approach that isn't flexible enough for others to work around or bypass.

Open to suggestions on what an API might look like with this in mind if you disagree.

Julienng:
If it can help, this is what I've done in the `onKeyDown` of `ComboboxInput` : 

```js
function onKeyDown(e) {
  if(!e.isDefaultPrevented()) {
    const container = popoverRef.current;
    if (!container) return;

    const element = container.querySelector("[aria-selected=true]");
    if (element) container.scrollTop = element.offsetTop - container.offsetTop;
  }
}
```
Another option is to use `scrollIntoView` but it didn't work well for my use case.

stephan281094:
> If it can help, this is what I've done in the `onKeyDown` of `ComboboxInput` :
> 
> ```js
> function onKeyDown(e) {
>   if(!e.isDefaultPrevented()) {
>     const container = popoverRef.current;
>     if (!container) return;
> 
>     const element = container.querySelector("[aria-selected=true]");
>     if (element) container.scrollTop = element.offsetTop - container.offsetTop;
>   }
> }
> ```
> 
> Another option is to use `scrollIntoView` but it didn't work well for my use case.

Thanks for this! I updated your code to make sure it:
- Only scrolls when selected element is partially in view.
- Keeps the selected element close to where it was (it doesn't scroll to the top).
- Takes care of the height of the selected element.
- Works when navigating from last item to first item and vice versa (requiring requestAnimationFrame).

```tsx
[...]

PR #383, opened 2019-11-20
comboboxoption scrolling issue resolved

prasad5795:
I have fixed the issue #357 hope it will resolve the problem

### DIFF
--- a/packages/combobox/src/index.js
+++ b/packages/combobox/src/index.js
@@ -541,6 +541,14 @@ export const ComboboxOption = forwardRef(function ComboboxOption(
 
   useEffect(() => {
     optionsRef.current.push(value);
+    const element = document.querySelector("li[data-highlighted]");
+    if (element) {
+      element.scrollIntoView({
+        behavior: "smooth",
+        block: "end",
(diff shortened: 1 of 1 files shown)


packages/combobox/src/index.js
```diff
@@ -541,6 +541,14 @@ export const ComboboxOption = forwardRef(function ComboboxOption(
 
   useEffect(() => {
     optionsRef.current.push(value);
+    const element = document.querySelector("li[data-highlighted]");
```
kiku511: @prasad5795 Wouldn't it be better to use a ref instead of querySelector?

chancestrickland:
Thank you for contributing! I'm going to move to close this PR for the time being. I've left some comments, but the biggest reason is that I think this is probably best handled in app code vs. the component itself.

There's no reason a developer shouldn't be able to measure refs and reactively scroll to fit specific app needs. If you'd like to try spinning up an example in the `examples` directory instead (taking some of my comments into account), that'd be helpful! And if there are any road blocks along the way I'd be happy to assist.

PR #386, opened 2019-11-20
Create @reach/descendants package

chancestrickland:


packages/descendants/README.md
```diff
@@ -0,0 +1,470 @@
...
+<Menu>
+  <CommonItems />
+  <Suspense>
+    <AsyncItems />
+  </Suspense>
+  <MenuItem />
+</Menu>
+```
+
+Now the indexes would be all out of whack. I _think_ when those `AsyncItems` render you'll lose the index on all the others, or get duplicates, I dunno.
```
diegohaz: Not sure if this will help with this particular case, but you can compare the element position in the DOM before adding it to the items array using [Node.compareDocumentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition). Example: https://github.com/reakit/reakit/blob/42e9dd0d39c53fae9661c83f470b919b45f7e845/packages/reakit/src/Rover/RoverState.ts#L141-L147

PR #387, opened 2019-11-21
[Tooltip] Make `ref` optional on `useTooltip` hook types

raunofreiberg:
Currently `ref` is required when calling the `useTooltip` hook. This shouldn't be the case, since `useTooltip` has its own internal ref, if not provided.

Related error:

image


raunofreiberg:
Ping @chancestrickland :-)

PR #391, opened 2019-11-24
[MenuButton] Ensure MenuPopover's collisions.bottom takes into account the button's height

grrowl:
This pull request:

---

This fixes a bug in MenuButton's MenuPopover where the menu is placed under the button, but the math for detecting whether it would collide with the bottom of the viewport doesn't take into account the button's height. 

This means the menu list will extend past the bottom of the page by as much as the menu button's height without detecting a collision.

### DIFF
--- a/packages/menu-button/src/index.js
+++ b/packages/menu-button/src/index.js
@@ -532,7 +532,8 @@ function getStyles(buttonRect, menuRect) {
   const collisions = {
     top: buttonRect.top - menuRect.height < 0,
     right: window.innerWidth < buttonRect.left + menuRect.width,
-    bottom: window.innerHeight < buttonRect.top + menuRect.height,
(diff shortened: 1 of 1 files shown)


chancestrickland:
Thanks @grrowl! I've got a PR open for MenuItem that replaces MenuPopover altogether with our generic @reach/popover package that should address this just as well. Hoping to get that merged this week so I don't think we'll need this ultimately, but I'll keep this open for the moment just in case.

grrowl:
either way! as a workaround we've added some really ugly padding to a viewport-bottom toolbar menu, so it'd be nice if either of the PRs could be released before next week

chancestrickland:
Problem addressed in https://github.com/reach/reach-ui/pull/397 and will be updated in the next release.

PR #397, opened 2019-12-01
@reach/menu-button: use hooks + descendants logic for better handling of focus

chancestrickland:
Knowing that a full rewrite of menu button is coming, this PR fixes a few long-standing problems in the interim. Mainly, I'm using the same logic we are currently using in Combobox to register focusable children rather than type checking components. Moving to hooks also removed some complexity from the top-level component around focus management without changing the behavior. 

Additionally, I added the ability to select menu items by spelling out an item's label as one does in native menus. I used the `valueText` prop label as a manual override here in case a menu item has multiple nested children with hierarchical text, but I did not want to document it until we've had the opportunity to discuss this approach and name of the prop. 

Last notable changes are 1) moving from `setState` to `useReducer`, and 2) fixing TS definitions around the `as` prop (which still feel a little odd, but they work well in the new example with Styled Components which was throwing errors before).


chancestrickland:
It's worth noting that using @reach/popover also resolves this issue on the RT site as well. https://github.com/ReactTraining/react-workshop-next/issues/9

PR #401, opened 2019-12-04
Add custom prop `position` to `menu-button`

kirillku:
Add custom prop `position` to `menu-button` similar to `tooltip`'s `position` prop.


chancestrickland:
Thanks for the PR! Just FYI, we will already address this in https://github.com/reach/reach-ui/pull/397 so I'm going to close this for the time being.

Issue #411, opened 2019-12-16
Tooltip configurable timers


skokenes:
### Current Behavior

The tooltip is hardcoded with the following timer values:

rest timer: 100ms
leave timer: 500ms

### Desired Behavior

These values sometimes produce an awkward UX, especially the leave timer of 500ms which can feel like an eternity after the user stops focusing an element with a tooltip. It would be great to be able to pass props to control what these timer values are.

### Suggested Solution

Provide props like `leaveTime`, `restTime` for the `Tooltip` component to override the 100ms, 500ms default values

### Who does this impact? Who is this for?

All users of the tooltip that want to customize it for their UX.

### Describe alternatives you've considered

My alternative is to fork the library - I don't see any way to accomplish the desired feature with the existing @reach API.

andreypopp:
The tooltip should probably be hidden immediately when element loses focus.

skokenes:
I believe it doesn't hide immediately when an element loses focus in order to have a better experience when moving focus from one element to another. A timer is used to prevent flashing a tooltip on and off.

My perspective is that this is great but the default time is too long, causing an awkward UX when going from focusing an element to not focusing anything. I'd like to shorten that time

chancestrickland:
We may consider this at some point but we're going to hold for the time being. Recommendations from W3 on specific tooltip behavior are still under consideration, and I suspect there may be spec around the timer details when they eventually land on something. I'd hate to introduce an API only to be forced to take it away.

patrickdbakke:
@chancestrickland I have this working on a local fork. Would you entertain a PR with experimental props?

```
<Tooltip enterTime={1000} leaveTime={1000}/>
```

I could include a `console.warn` message "`enterTime` and `leaveTime` are non-standard props for tooltip timing, and this api may change at any tme"

chancestrickland:
Per my comment above, I would prefer not to do that at this time. We may consider it after our WebAIM review, but in the mean time If you have a need for this I'd recommend using https://www.npmjs.com/package/patch-package

patrickdbakke:
Sounds good! Thanks @chancestrickland !

gnapse:
Isn't it the current recommendation to remove the tooltip as soon as the element looses focus or on mouse out?

> If the tooltip is invoked when the trigger element receives focus, then it is dismissed when it no longer has focus (onBlur). If the tooltip is invoked with mouseIn, then it is dismissed with on mouseOut.

[...]

PR #417, opened 2020-01-03
@reach/tooltip - Change leaving time

eng-almeida:

This pull request:

- [ ] Creates a new package
- [x] Other

If creating a new package:

  - [ ] `examples` directory

## Tooltip: Update leaving time

The intent of this PR is to standardize the tooltip visible leaving time with the [entering time](https://github.com/reach/reach-ui/blob/master/packages/tooltip/src/index.js#L201).

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -210,7 +210,7 @@ let leavingVisibleTimer;
 
 function startLeavingVisibleTimer() {
   clearTimeout(leavingVisibleTimer);
-  leavingVisibleTimer = setTimeout(() => transition("timecomplete"), 500);
+  leavingVisibleTimer = setTimeout(() => transition("timecomplete"), 100);
 }
 
 function clearLeavingVisibleTimer() {


chancestrickland:
@eng-almeida Can you explain your rationale a bit more? What is the argument over choosing a 100ms delay vs. a 500ms delay? Is this an arbitrary preference or is there precedent for this choice that we've missed?

eng-almeida:
@chancestrickland I should have put a better description on the PR description, sorry about that.
As it is right now there are 2 different timer values, 100 ms for the enter event and 500 ms for the leave event. Is this the intended behaviour? Wouldn't make sense to use the same value? Also isn't 500 ms a bit too much?

chancestrickland:
Personally I feel like it makes sense to have the tooltip stick around a bit longer on exit. If you don't quite rest the cursor steadily at first, it gives the user a small opportunity to correct without seeing a flicker. There doesn't appear to be a standard here, and I'm opening to exposing props for a bit more control here.

chancestrickland:
I'm going to close this for now per my comment here: https://github.com/reach/reach-ui/issues/411#issuecomment-573970811

We may revisit this after the 1.0 release and a subsequent audit from WebAIM, but I don't want to introduce any changes here without better guidance on the accessibility impact.

PR #430, opened 2020-01-15
Allow passing in an optional name prop for <Combobox />

et:
I am using [react-hook-form](https://react-hook-form.com/) as my form library. This form library expects a way to pass a`ref` in which `ComboboxInput` already provides. Additionally, it also expects that the internal DOM element has a `name` prop defined which `ComboInput` doesn't have.

Hence, this PR allows an optional `name` prop that gets passed to the underlying `<input />` tag.

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -402,6 +402,7 @@ export const ComboboxInput = forwardRefWithAs<"div", ComboboxInputProps>(
       onBlur,
       onFocus,
       value: controlledValue,
+      name,
       ...props
     },
     forwardedRef
@@ -489,6 +490,7 @@ export const ComboboxInput = forwardRefWithAs<"div", ComboboxInputProps>(
         {...props}
         data-reach-combobox-input=""
         ref={ref}
+        name={name}
         value={inputValue || ""}
         onClick={wrapEvent(onClick, handleClick)}
         onBlur={wrapEvent(onBlur, handleBlur)}
@@ -538,6 +540,10 @@ export type ComboboxInputProps = {
    * @see Docs https://reacttraining.com/reach-ui/combobox#comboboxinput-value
    */
   value?: ComboboxValue;
+  /**
+   * The name attribute of the internal <input /> element.
+   */
+  name?: string;
 };
 
 ComboboxInput.displayName = "ComboboxInput";


packages/combobox/src/index.tsx
```diff
@@ -489,6 +490,7 @@ export const ComboboxInput = forwardRefWithAs<"div", ComboboxInputProps>(
         {...props}
```
chancestrickland: Note here we spread all of the other props into the underlying input

chancestrickland:
This shouldn't be necessary. You can already pass a name prop, as any props not explicitly defined are spread into the underlying input component. I left a comment in the code you submitted for reference.

chancestrickland:
Curious though, are you getting a type error when attempting to use a name prop? If so I think I may need to make an adjustment elsewhere.

et:
Thanks for the response @chancestrickland. I am getting the following error:

```ts
<ComboboxInput name="text" />
```
Screen Shot 2020-01-16 at 1 31 23 PM


```ts
No overload matches this call.
  Overload 1 of 2, '(props: ComboboxInputProps & Pick<any, string | number | symbol> & { as?: any; } & { as: any; }): Element', gave the following error.
    Property 'as' is missing in type '{ name: string; }' but required in type '{ as: any; }'.
  Overload 2 of 2, '(props: PropsWithAs<"div", ComboboxInputProps>): Element', gave the following error.
    Type '{ name: string; }' is not assignable to type 'IntrinsicAttributes & ComboboxInputProps & Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "placeholder" | ... 250 more ... | "onTransitionEndCapture"> & { ...; }, "ref" | ... 251 more ... | "onTransitionEndCapture"> & { ...; }'.
      Property 'name' does not exist on type 'IntrinsicAttributes & ComboboxInputProps & Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "placeholder" | ... 250 more ... | "onTransitionEndCapture"> & { ...; }, "ref" | ... 251 more ... | "onTransitionEndCapture"> & { ...; }'.ts(2769)
types.d.ts(14, 9): 'as' is declared here.
```

PR #432, opened 2020-01-19
[Several Packages] Remove several type casts

skovy:

This pull request:

- [ ] Creates a new package
- [x] Other (types)

The inspiration for these changes [was this tweet](https://twitter.com/chancethedev/status/1218399362746933248) 😅. I didn't see any contributing guidelines _(I could've missed them)_ so I wasn't confident if it was necessary to do a PR per package? If that's the case, I'm happy to separate out these changes!

I searched for `as any` in the codebase and attempted to resolve most of the more straightforward ones. There are a few more but those will likely be much more involved. 

There were a few types of changes:

- Converted the prop types `() => {}` to `() => null` to comply with their type definitions. [They expect either `Error` or `null`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d459d36f4e1f47c3c6273b287718c79426e41a9f/types/prop-types/index.d.ts#L41).
- Converted the `ElementTypes`'s `RegionTypes`'s value to have a index signature of a `string`. This allowed removing the type case of `key as any` and to complies with `Object.keys`: ["An array of strings that represent all the enumerable properties of the given object."](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Syntax).
```js
// array like object
const obj = { 0: 'a', 1: 'b', 2: 'c' };
console.log(Object.keys(obj)); // console: ['0', '1', '2']
```
- Simplified `isRightClick`. Previously the statement `"which" in nativeEvent` was evaluated as always true so the second `else if` was always `never`. If accessing one of these fields and it doesn't exist it should be `undefined` and since this was using triple equals we should be safe against something like `0 == undefined`.  So it should be safe to remove the check before accessing the fields? I'm happy to revert if this is an unsafe change. I verified with logic like this: 
```js
const ev = new MouseEvent("myDummyTest")
ev.which === 1 || ev.button === 1 // true
ev.wat === 1 || ev.idk === 1 // false
```
- Since it appears `currentStyle` was used very intentionally for IE support ([assuming from the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/currentStyle)) I added a type `HTMLElementWithCurrentStyle` to allow accessing that attribute since it will probably never get added to the TS libraries.
- Allowed `useForkedRef` to accept `undefined` because it should safely handle in `assignRef` with the _double_ equality check of `if (ref == null) return;`.
- Fixed unrelated typo

### DIFF
--- a/packages/alert-dialog/src/index.tsx
+++ b/packages/alert-dialog/src/index.tsx
@@ -78,7 +78,7 @@ export const AlertDialogOverlay = forwardRef<HTMLDivElement, AlertDialogProps>(
       >
         <DialogOverlay
           {...props}
-          ref={forwardRef as any}
+          ref={forwardRef}
           data-reach-alert-dialog-overlay
           initialFocusRef={leastDestructiveRef}
         />
@@ -138,7 +138,7 @@ export const AlertDialogContent = forwardRef<
       role="alertdialog"
       aria-labelledby={labelId}
       {...props}
-      ref={forwardRef as any}
+      ref={forwardRef}
       // lol: remove in 1.0
       data-reach-alert-dialong-content
       data-reach-alert-dialog-content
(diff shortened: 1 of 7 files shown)


chancestrickland:
Beautiful. Thanks so much for the assist here! I'll take a look at everything tomorrow and let you know if I have any questions or concerns before merging.

Issue #149, opened 2019-04-23
Dialog: Allow manual control of focus after dialog closes


brendancarney:
Dialog uses FocusLock, which returns focus to the last element that had the focus when the dialog is closed. In our case, we actually want to focus something else. Currently we have to work around this by using `setTimeout` to make sure our focus code runs after Reach UIs. 

What do you think about adding support for controlling this focus behavior? I'm not sure exactly what it would look like. A couple options:

- A prop (`returnFocus` maybe?) to disable the auto focusing. 
- Some function passed to `onDismiss` that allows prevention of the auto focusing. 

Thanks!

briandipalma:
I'm happy to create a PR if we get the go ahead for this as we're also looking for this feature.

briandipalma:
This is the workaround I used in case others come across this issue. It makes me thing the cleanest API would be for `returnFocus` to be supported by the Dialog etc components.

```javascript
function blurActiveElement() {
  setTimeout(function() {
    document.activeElement.blur();
  }, 0);
}

export default function PopupDialog({
  onDismiss,
  isOpen,
  returnFocus,
  children
}) {
  // This function is run either on component unmount or when the two dependent
  // values change.
  // If when the effect was created the Dialog was open and not meant to return
  // focus then we blur the active element. Strictly speaking it would be wrong
  // to run it if only `returnFocus` changed but as the Dialog is still open in
  // that case I see no issue.
  useEffect(
    () => () => {
      if (returnFocus === false && isOpen === true) {
        blurActiveElement();
      }
    },
    [returnFocus, isOpen]
  );

  ...
}
```

mdugue:
Any news here? I really enjoy using reach-dialog with it's great details. For me the only missing part is better control on the focus return upon closing it. 😅I can image quite some use cases potentially relevant to many people where you want to opt out of the default behavior. Would you be open to a PR adding a `disableAutomaticFocusReturn` prop? Any recommendations on that wording / API surface?

Thanks a lot in advance!!

mdugue:
actually, I just discovered this PR https://github.com/reach/reach-ui/pull/450. @chancestrickland  any chances that this could get merged?

chancestrickland:
I understand why this might be necessary, but I do want to consider the API a bit more before merging #450. Keep in mind that you can always patch this behavior in the mean time if you'd like with https://www.npmjs.com/package/patch-package

mdugue:
Sure! Any API ideas you want to share and bounce?

[...]

PR #450, opened 2020-01-31
Manual control for focus return for Dialog

jromerof:
This pull request:
- [X] Adds additional features/functionality to an existing package

Hi! 
First of all this is my first time opening a PR in a project that is not mine so if I screwed up something feel free to delete, also I hope it complies to the PR rules. Also I'm a total newbie at TS.

In this PR I added a ``returnFocusRef`` prop to Dialog and a focus return example. 

The ``returnFocusRef`` gives focus to the element passed in the ref by turning returnFocus to false and passing a ``onDeactivation`` function to the Focus Lock component (as per the [FocusLock docs](https://github.com/theKashey/react-focus-lock#unmounting-and-focus-management) )

The example is called 'Return Focus'  it should begin with the modal open and on close focus the button that don't open the dialog.

This could be a potential solution to [this issue](https://github.com/reach/reach-ui/issues/149)

Also thanks to the team for reach/ui I love it!!

### DIFF
--- a/packages/dialog/src/index.tsx
+++ b/packages/dialog/src/index.tsx
@@ -64,6 +64,7 @@ const DialogInner = forwardRef<HTMLDivElement, DialogProps>(
     {
       allowPinchZoom,
       initialFocusRef,
+      returnFocusRef,
       onClick,
       onDismiss = noop,
       onMouseDown,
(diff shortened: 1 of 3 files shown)


mdugue:
I would love to see this (or a similar approach) merged

packages/dialog/src/index.tsx
```diff
@@ -75,13 +76,21 @@ const DialogInner = forwardRef<HTMLDivElement, DialogProps>(
...
 
     const activateFocusLock = useCallback(() => {
       if (initialFocusRef && initialFocusRef.current) {
         initialFocusRef.current.focus();
       }
     }, [initialFocusRef]);
 
+    const deactivateFocusLock = useCallback(() => {
+      if (returnFocusRef && returnFocusRef.current) {
+        setTimeout(() => returnFocusRef.current.focus(), 0);
```
indiesquidge, line 90: It may be worth leaving a comment on why the zero-timeout is needed here. What do you think about either adding a comment like "Without the zero-timeout, focus will likely remain on the button/control" or [linking to the FocusLock docs directly][1]? I like the idea of linking to the docs more.

[1]: https://github.com/theKashey/react-focus-lock#unmounting-and-focus-management

packages/dialog/src/index.tsx
```diff
@@ -253,6 +268,10 @@ export type DialogProps = {
    * @see Docs https://reacttraining.com/reach-ui/dialog#dialogoverlay-initialfocusref
    */
   initialFocusRef?: React.RefObject<any>;
+  /**
+   * This ref can help you to override the returnFocus of the FocusLock component
+   */
```
indiesquidge, line 273: What do you think about making this doc comment more similar to the `initialFocusRef` comment, since they are essentially inverses of each other?

"By default the element that opened the dialog will receive focus when the dialog closes, but you can provide a ref to focus instead."

chancestrickland:
Can you please open a new PR to the `develop` branch? Thanks!

techjpiron:
Hi there!

Is there any chance that this feature will be merge? If there is help needed I'm willing to help out!

@indiesquidge @chaance

PR #463, opened 2020-02-11
New component: @reach/radio

chancestrickland:
This is a work in progress. The component code itself is ready for review. Working on tests, docs, and a few more examples.

This pull request:

- [x] Creates a new package
- [ ] Other

If creating a new package:

  - [x] `examples` directory
  - [x] `src` directory with an `index.tsx` entry file


packages/radio/README.md
```diff
@@ -0,0 +1,7 @@
+# @reach/tabs
```
hockeybuggy: I think this should say `@reach/radio`?

hockeybuggy: (sorry not sure if this is helpful or mostly annoying since this component is WIP)

packages/radio/examples/example-styles.css
```diff
@@ -0,0 +1,90 @@
+:root {
```
raunofreiberg, line 1: Should probably move these to `@reach/radio/styles.css` to provide some default styles.

packages/radio/src/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+      defaultValue,
+      disabled,
+      value: valueProp,
+      name,
+      onChange,
+      ...props
+    },
+    forwardedRef
+  ) {
+    let { current: isControlled } = useRef(typeof valueProp !== "undefined");
```
raunofreiberg, line 62: Let's warn about switching between controlled/uncontrolled behavior?

packages/radio/src/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+      type: RadioGroupEvents.Select,
+      callback: onChange,
+      disabled,
+      domEvent: event.nativeEvent,
+      node: ownRef.current,
+      value,
+    });
+  }
+
+  function handleKeyDown(event: React.KeyboardEvent) {
```
raunofreiberg, line 249: I know it didn't exist at the time of this PR, but what do you think about using the [`useDescendantKeydown`](https://github.com/reach/reach-ui/blob/52d721cd643d1a6d24c7253ab24e114be262d092/packages/descendants/src/index.tsx#L221) abstraction here?

packages/radio/src/machine.ts
```diff
@@ -0,0 +1,363 @@
...
+  };
+}
+
+export type MachineToReactRefMap<TE extends MachineEventWithRefs> = {
+  [K in keyof TE["refs"]]: React.RefObject<TE["refs"][K]>;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+export function useMachine<
```
raunofreiberg, line 310: We can use `@reach/machine` instead of this now! 🎉

packages/radio/src/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+      isControlled,
+      isRTL,
+      name,
+      onChange,
+      send,
+      state: current,
+      value: current.context.value,
+    };
+
+    if (isControlled && current.context.value !== valueProp) {
```
raunofreiberg, line 95: Should we do this in render? Would `useEffect` be better?

chancestrickland:
@raunofreiberg Appreciate the feedback! We're probably going to press pause on releasing new components for the moment and focus on fixing bugs elsewhere. Keeping this PR open until we're ready to revisit it.

jaredpalmer:
@raunofreiberg i'm not maintaining anymore, but a similar accessible component lives at https://github.com/palmerhq/radio-group

tonypee:
@jaredpalmer - thanks for the link - we'll look at using that until reach gets a radio! seems an important thing to have

Issue #467, opened 2020-02-13
[Combox] Missing aria connection between ComboboxInput and ComboboxPopover


sebpettersson:
Looking at the [examples](https://reacttraining.com/reach-ui/combobox), this doesn't look correct:

```
<div aria-haspopup="listbox" aria-owns="listbox--268" aria-expanded="true" data-reach-combobox="" role="combobox">
  <input aria-autocomplete="both" aria-labelledby="demo" data-reach-combobox-input="" id="listbox--268" value="ap">
</div>

<div data-reach-popover="" data-reach-combobox-popover="" tabindex="-1" style="position: absolute; width: 173px; left: 351px; top: 1908px;">
  <ul aria-labelledby="demo" data-reach-combobox-list="" role="listbox"></ul>
</div>
```
There's no explicit connection between the input field and the popover (except sharing `aria-labelledby`), and `aria-owns` on `role="combobox"` points to the `input` field which doesn't make sense since it's already a child element in the DOM.

In contrast, the markup for the [Combobox example on w3.org](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html) looks like this:

```
  <div role="combobox" aria-expanded="true" aria-owns="ex1-listbox" aria-haspopup="listbox" id="ex1-combobox">
    <input type="text" aria-autocomplete="list" aria-controls="ex1-listbox" id="ex1-input" aria-activedescendant="">
  </div>

  <ul aria-labelledby="ex1-label" role="listbox" id="ex1-listbox" class="listbox"></ul>
```
Here, `aria-owns` points to the the actual listbox instead (which is correct since it's not a child element in the DOM), and additionally, the `input` has `aria-controls` which points to the same.

sebpettersson:
Fixed by #468

PR #468, opened 2020-02-13
Fix WAI-ARIA attributes in Combobox

sebpettersson:
Potential fix for #467

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -498,6 +498,7 @@ export const ComboboxInput = forwardRefWithAs<ComboboxInputProps, "input">(
         aria-activedescendant={
           navigationValue ? String(makeHash(navigationValue)) : undefined
         }
+        aria-controls={listboxId}
         {...props}
         data-reach-combobox-input=""
         ref={ref}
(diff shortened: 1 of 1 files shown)


packages/combobox/src/index.tsx
```diff
@@ -869,7 +870,6 @@ export const ComboboxButton = forwardRefWithAs<{}, "button">(
 
     return (
       <Comp
-        aria-controls={listboxId}
```
chancestrickland: I believe we should leave this, even though it isn't in the ARIA example. We can probably make some other improvements to the button specifically for screen readers, but so long as it's here and controls the popover we should keep the aria context.

sebpettersson: Sorry about that, I actually misread the code, it's fixed now.

Though now I'm unsure how ComboboxButton is used - it's not mentioned in the docs?

chancestrickland: There's an example in the repo that shows how it's used, though I do need to think through its implementation a bit before we add documentation for it. The challenge AFAICT is that its functionality changes a bit depending on whether or not there is anything to render inside the popover, and that depends on how/whether or not your app decides to render a popover when matching options are not found. I'll have to chat with Ryan about some of these details to get some clarity before documenting that component and making it "official".

sebpettersson: Got it, thanks for the explanation.

PR #486, opened 2020-02-21
tabs: Add support for vertical orientation

peterjmag:
Hey folks 👋

This PR adds a new `orientation` prop to the Tabs component and sets its default to `"horizontal"`. The new prop has these effects:

- Adds the `aria-orientation` attribute to the TabsList DOM element per [the WAI-ARIA docs](https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-20).
  - Since `orientation` has a default, that means the existing tab examples also get this new attribute.
- When it's set to `"vertical"`:
  -  Changes the flex-direction of TabsList to stack tabs vertically.
  - Displays TabPanels to one side.
  - Allows the user to cycle through tabs with the up + down arrow keys (instead of the left + right arrows).
  - Allows the user to focus the active TabPanel with the right arrow key (or the left arrow key in RTL mode).

I did see this TODO:

https://github.com/reach/reach-ui/blob/9bc8e3eb0da2ed977c5b9736e0ac3483d65a96de/packages/tabs/src/index.tsx#L17-L21

This PR takes a slightly different approach, by limiting the options to just "horizontal" and "vertical" and then relying on RTL (or just the ordering of TabsList and TabPanels) for the rest. Let me know what you think!

I'm opening this as a draft PR, because I'm still trying to figure out what's going on with the tests, but I wanted to get some feedback on the approach so far.

## Screenshots

Screen Shot 2020-02-21 at 13 54 22

---

Screen Shot 2020-02-21 at 13 54 38

## PR template

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory


packages/tabs/src/index.tsx
```diff
@@ -279,14 +296,20 @@ export const TabList = forwardRefWithAs<TabListProps, "div">(function TabList(
 
   let handleKeyDown = wrapEvent(
     function(event: React.KeyboardEvent) {
-      if (event.key === "ArrowDown") {
+      if (
+        orientation === "vertical"
+          ? isRTL
+            ? event.key === "ArrowLeft"
+            : event.key === "ArrowRight"
```
peterjmag: One thing I'm just realizing: this doesn't take into account the relative order of TabsList and TabPanels. For instance:

```jsx
<Tabs orientation="vertical">
  <TabPanels>
    <TabPanel>
      <h1>one!</h1>
      <p>Here's some example content.</p>
      <button>yo</button>
    </TabPanel>
    <TabPanel>
      <h1>two!</h1>
      <p>Here's some example content.</p>
    </TabPanel>
    <TabPanel>
      <h1>three!</h1>
      <p>Here's some example content.</p>
    </TabPanel>
  </TabPanels>

  <TabList>
    <Tab>One</Tab>
    <Tab>Two</Tab>
    <Tab>Three</Tab>
  </TabList>
</Tabs>
```

Assuming we're in LTR mode, TabPanels will show up on the left (which is correct), but the left arrow key won't focus the panel.

I'll think about how to address that, but let me know if you all have any ideas. Could actually be a good reason to make the options more explicit as @chancestrickland mentioned in his TODO i.e. `horizontal-top`, `horizontal-bottom`, etc.

website/src/pages/tabs.mdx
```diff
@@ -143,7 +144,7 @@ function Example() {
       onChange={index => setTabIndex(index)}
       style={{
         color: "white",
-        background: backgroundColor
+        background: backgroundColor,
```
peterjmag: Oops, my local Prettier config might have interfered here. I'll revert these comma changes.

website/src/pages/tabs.mdx
```diff
@@ -91,6 +91,7 @@ Please see the [styling guide](/styling).
 | [onChange](#tabs-onchange)         | function      | false    |
 | [defaultIndex](#tabs-defaultindex) | number        | false    |
 | [index](#tabs-index)               | number        | false    |
+| [orientation](#tabs-orientation)   | string        | false    |
```
peterjmag: Is `string` enough for the type here? Or should I make it more specific, a la [line 250](https://github.com/reach/reach-ui/pull/486/files#diff-2db3c8b9bcec3e34f466572fcfd6df2bR250)?

packages/tabs/__tests__/tabs.test.tsx
```diff
@@ -51,6 +51,56 @@ describe("<Tabs />", () => {
...
+        `#${container
+          .querySelector(`.${buttonId}`)!
+          .getAttribute("aria-controls")}`
+      );
+    }
+
+    fireEvent.click(firstTab);
+
+    fireEvent.keyDown(tabList, { key: "ArrowDown", code: 40 });
+    expect(document.activeElement).toBe(container.querySelector(".tab-2"));
```
peterjmag: Strangely, I'm getting different results from [what I'm getting from Travis](https://travis-ci.com/reach/reach-ui/builds/150023686) when I run `yarn test` locally. For me, the tests fail here on line 78, with this output:
https://gist.github.com/peterjmag/cfd150fb92a04635c5d1a58921edd197

Digging into this too.

chancestrickland:
Looks like a solid start, thanks for the PR! I'll dig into this a bit next week and see what we can do.

packages/tabs/__tests__/tabs.test.tsx
peterjmag: Turns out I had some outdated built files in dist/ when I last updated the snapshots (and the tests run against those built files). I re-ran the build and then updated the snapshots again in https://github.com/reach/reach-ui/pull/486/commits/b10530560716985c963321a79e54500d6d8e31e9, but now I'm getting differing IDs from Travis:

```
              <button
    -           aria-controls="tabs--15--panel--0"
    +           aria-controls="tabs--REACH_ID_--panel--0"
                aria-selected="true"
                class="tab-1"
                data-reach-tab=""
                data-selected=""
    -           id="tabs--15--tab--0"
    +           id="tabs--REACH_ID_--tab--0"
                role="tab"
                tabindex="0"
              >
```
https://travis-ci.com/reach/reach-ui/builds/150269550#L431

packages/tabs/src/index.tsx
peterjmag: I took a shot at detecting the relative order (while still taking isRTL into account) in https://github.com/reach/reach-ui/pull/486/commits/f07be1d61eb10ce2c718cba6d0cd2704efb563a4.

packages/tabs/__tests__/tabs.test.tsx
peterjmag: Ended up just adding an ID to the test component itself to satisfy Travis: 
https://github.com/reach/reach-ui/pull/486/commits/953169dc466a0f8f934d0bfc6d45f19a90f839c1#diff-a65e1b1f6a8ab0f93d7a3e2afee4d44eR138

peterjmag:
Hey @chancestrickland! The tests are now running through, so this PR is (finally) ready for a proper review whenever you get a chance ☺️

chancestrickland:
Thanks so much for this! Finally got around to reviewing, and it looks great. I made some adjustments to the API. Mainly, `orientation` will accept any of the following:
 - `horizontal-start` (default)
 - `horizontal-end`
 - `vertical-start`
 - `vertical-end`

The reason here is that I'd rather be explicit vs. checking children to see whether TabList or TabPanels comes first, which can break certain composition patterns. We could register refs similar to how we do for tab descendants, but a single prop change at the top along with added documentation to explain the distinctions seems simple enough for this purpose. App code can abstract this further if someone doesn't like it.

chancestrickland:
Slept on it, and now I'm having second thoughts. I'm going to seek some feedback on this approach. `direction` and `orientation` are two separate things at the end of the day, and we might be making too many assumptions.

peterjmag:
@chancestrickland No problem, and definitely no rush given everything that's going on right now. In any case, let me know if there's anything I can do to help gather feedback!

Issue #472, opened 2020-02-14
[Accordion] Animation


raunofreiberg:
## ❓Question

I'm having a hard time animating Accordion properly with height. Accordion panel being set to `display: none` is not helping to achieve an animation either. Would really appreciate an example or some help 🙂

raunofreiberg:
I could get it to work by using `react-spring`, controlling the index state of Accordion, and passing `isActive` to each `AccordionItem` by mapping over the children of `Accordion`.

Although:

1) I had to set `display: block !important` to `AccordionPanel` to make `react-spring` exit animation work.
2) Can I do this without controlling state? I would very much like to have the same result, but keep the `multiple` and `collapsed` props working.

![gif](https://i.gyazo.com/3bd24fe03ca2e51ef2583cbf379d9a72.gif)

raunofreiberg:
@chancestrickland Would it be too much of a "hack" to pass `isActive` to `AccordionItem` by Reach [here](https://github.com/reach/reach-ui/blob/master/packages/accordion/src/index.tsx#L394)? I see that Tabs has the same logic, but it seems like this is up for deprecation?

My thinking here is that I don't want to handle the `collapsed` and `multiple` cases myself, and would rather just Reach tell me whether my AccordionItem is active.

raunofreiberg:
Friendly ping @chancestrickland 

I think implementing animations is quite a common case for accessible components. Would love to come to a solution that doesn't affect the functionality 🙃

chancestrickland:
On my todo list! May not get to it today but should be able to take a look next week.

PR #500, opened 2020-03-05
feat(Accordion): expose accordion item state for animations

raunofreiberg:
Hey 👋

This PR is to enable animating Accordion without controlling the state, which would end up duplicating a lot of the `collapsible` and `multiple` logic already included. I think one should not have to duplicate functionality to achieve animations 🙂

To make animating panels possible without modifying functionality, I'm exporting the `AccordionItemContext` which has all the necessary properties to be used within a custom `AnimatedAccordionPanel` component.

re: https://github.com/reach/reach-ui/issues/472#issuecomment-586336594

gif

<hr />

I'm happy to hear any other approaches or thoughts on this 🙂

### DIFF
--- a/packages/accordion/examples/animated.example.js
+++ b/packages/accordion/examples/animated.example.js
@@ -1,0 +1,93 @@
+import React, { useRef, useEffect, useState, forwardRef } from "react";
+import {
+  Accordion,
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel,
+  useAccordionItemContext,
+} from "@reach/accordion";
+import { animated, config, useSpring } from "react-spring";
+import { action } from "@storybook/addon-actions";
+import "@reach/accordion/styles.css";
+
+let name = "Animated";
+
+function useDivHeight() {
+  const ref = useRef(null);
+  const [height, setHeight] = useState(0);
+
+  useEffect(() => {
+    const resizeObserver = new ResizeObserver(([entry]) => {
+      requestAnimationFrame(() => {
+        if (!entry) {
+          return;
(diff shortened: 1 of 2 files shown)


raunofreiberg:
cc @chancestrickland

packages/accordion/examples/animated.example.js
```diff
@@ -0,0 +1,93 @@
...
+  const { ref, height } = useDivHeight();
+  const animation = useSpring({
+    opacity: open ? 1 : 0,
+    height: open ? height : 0,
+    overflow: "hidden",
+    config: config.default,
+  });
+
+  return (
+    <AnimatedAccordionPanel style={animation} hidden={false} ref={forwardedRef}>
```
raunofreiberg: I had to set `hidden` to `false` here because the base `AccordionPanel` component sets it to `true` when not open.

chancestrickland:
Thanks for the PR! Closing in favor of e93835686cec6cf4c25e679629cbd6e4e4ceb838, I am adding context hooks for several components that follow this pattern. We want to expose some state to help with these sort of things without barfing out all of our internal context.

packages/accordion/examples/animated.example.js
chancestrickland: Just FYI, when you do this you should then set `aria-hidden` to true when the panel is collapsed so that it is hidden from the accessibility tree.

chancestrickland:
BTW, I did adopt the animation example. Should have still merged the PR and made changes on top, but I added the hook before I remember to check back. My bad there!

PR #501, opened 2020-03-06
[Combobox] Add popover visibility hook for styling

indiesquidge:
Influenced by the [Reach UI Philosophy][1] on exposing state for JS styling, and by the [recent addition of useTabsContext][2].

```jsx
const StyledComboboxInput = (props) => {
  const { isExpanded } = useComboboxContext();

  return (
    <ComboboxInput
      {...props}
      style={{ backgroundColor: isExpanded ? "blue" : "red" }}
    />
  );
};
```

The real-world desire for this feature comes from a design I received that changes the inputs bottom border radii when the popover is open. You can see I tried to implement it based on input focus and value change, but this doesn't handle every case and adds unnecessary state and complexity.

Before hook

After hook

---

[1]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#exposing-state-for-js-styling
[2]: https://github.com/reach/reach-ui/blob/d4bcfc73f27e6e618ec2eb2b0154a58bc17496f0/packages/tabs/src/index.tsx#L685-L695

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -746,7 +746,6 @@ export const ComboboxOption = forwardRefWithAs<ComboboxOptionProps, "li">(
           // onBlur will work as intended
           tabIndex={-1}
           onClick={wrapEvent(onClick, handleClick)}
-          // @ts-ignore
           children={children || <ComboboxOptionText />}
         />
       </OptionContext.Provider>
@@ -1158,11 +1157,27 @@ export function escapeRegexp(str: string) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+export function useComboboxContext(): ComboboxContextValue {
+  let { isVisible: isExpanded } = useContext(ComboboxContext);
+  return useMemo(
+    () => ({
+      isExpanded,
+    }),
+    [isExpanded]
+  );
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 // Well alright, you made it all the way here to like 1100 lines of code (geez,
 // what the heck?). Have a great day :D
 
 ////////////////////////////////////////////////////////////////////////////////
 // Types
+
+export type ComboboxContextValue = {
+  isExpanded: boolean;
+};
 
 type DescendantProps = {
   value: ComboboxValue;
--- a/packages/combobox/examples/using-popover-visibility-hook.example.tsx
+++ b/packages/combobox/examples/using-popover-visibility-hook.example.tsx
@@ -1,0 +1,62 @@
+import React from "react";
+import {
+  Combobox,
+  ComboboxInput,
+  ComboboxList,
+  ComboboxOption,
+  ComboboxPopover,
+  ComboboxInputProps,
+  useComboboxContext,
+} from "@reach/combobox";
+import "@reach/combobox/styles.css";
+
+let name = "Using Popover Visibility Hook (TS)";
+
+const StyledComboboxInput = (props: ComboboxInputProps) => {
+  const { isExpanded } = useComboboxContext();
+
+  return (
+    <ComboboxInput
+      {...props}
+      style={{
+        width: 400,
+        fontSize: "100%",
+        padding: "0.33rem",
+        backgroundColor: isExpanded ? "cornsilk" : "aliceblue",
+      }}
+    />
+  );
+};
+
+function Example() {
+  return (
+    <div>
+      <h2 id="choose-a-fruit">Choose a fruit</h2>
+      <p>Input styles change based on popover visibility!</p>
+      <Combobox>
+        <StyledComboboxInput aria-labelledby="choose-a-fruit" />
+        <ComboboxPopover style={popupStyle}>
+          <ComboboxList>
+            <ComboboxOption value="Apple" />
+            <ComboboxOption value="Banana" />
+            <ComboboxOption value="Orange" />
+            <ComboboxOption value="Pineapple" />
+            <ComboboxOption value="Kiwi" />
+            <button>Hi</button>
+          </ComboboxList>
+        </ComboboxPopover>
+      </Combobox>
+    </div>
+  );
+}
+
+Example.story = { name };
+export const Comp = Example;
+export default { title: "Combobox" };
+
+////////////////////////////////////////////////////////////////////////////////
+
+const popupStyle = {
+  boxShadow: "0px 2px 6px hsla(0, 0%, 0%, 0.15)",
+  border: "none",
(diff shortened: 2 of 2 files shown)


packages/combobox/src/index.tsx
```diff
@@ -250,6 +250,10 @@ const ComboboxContext = createNamedContext(
   {} as IComboboxContext
 );
 
+export function useIsComboboxPopoverVisible() {
+  return useContext(ComboboxContext).isVisible;
+}
+
```
indiesquidge: We could export `ComboboxContext` as a whole, but I'm opting for exporting individual "slices" of context for a few reasons:

- This is the desired way based on the [Reach UI Philosophies][1]
- [Other libraries][2] follow this same pattern
- It allows us to be more conservative with what parts of context we're okay exposing

[1]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#exposing-state-for-js-styling
[2]: https://github.com/ReactTraining/react-router/blob/b77283cb75efc909c34e48c7cdcb7f50281f4055/packages/react-router/modules/hooks.js#L9

chancestrickland: Your instinct is correct, IMO. That said, this hook feels a little less than ergonomic. Curious if we can't establish a better pattern for the library as a whole where we expose bits of state folks should reasonably expect. This would be great for pretty much all components with an indexed collection of descendants.

indiesquidge: Hmm, I'm not sure I follow. Can you help me understand why exposing this slice of context is not ergonomic? Is there a different export you'd be more comfortable with? What do you think is the best actionable step forward?

indiesquidge:
@chancestrickland if we're uncomfortable with this PR, do you have advice on how else to handle styling based on listbox visibility?

chancestrickland:
@indiesquidge I merged a change recently where we expose its visibility as an argumet in the render prop. I definitely want have some hooks available as well. Bear with me, unfortunately I'm not able to work on this full time anymore so I may be a bit slower getting to PRs.

indiesquidge:
> Bear with me, unfortunately I'm not able to work on this full time anymore so I may be a bit slower getting to PRs.

@chancestrickland, no worries there, you've been incredibly generous with your time and effort maintaining this library and responding to community requests and issues. I was sad to hear about the effect the pandemic had on React Training.

If you would like any assistance maintaining this library, pushing releases, triaging issues, etc., I'd be more than happy to help. Reach UI means a lot to me; I'd like to give back in whatever way I can. 💙

---

I saw you added a change to Tabs to expose context for some of the internal state 

https://github.com/reach/reach-ui/blob/d4bcfc73f27e6e618ec2eb2b0154a58bc17496f0/packages/tabs/src/index.tsx#L685-L695

Would you prefer I implement the hook in this PR in a similar way?

```jsx
 export function useComboboxContext() { 
   let { isVisible } = useContext(ComboboxContext); 
   return useMemo( 
     () => ({ 
       isVisible,
     }), 
     [isVisible] 
   ); 
 } 
```

chancestrickland:
@indiesquidge Yes, the goal of that addition to tabs was to come up with a standard way to expose a limited portion of internal state via `use{COMPONENT}Context`. I want to start small – basic states that might be used for styling, generated IDs that might be useful for nested aria. We can introduce more if we need it, but it'll be harder to reduce the API surface later if we change our minds. If you want to update this PR with that in mind, I'll be happy to merge. 

As far as your offer to help more broadly, gladly accepted. Let me get through this week, but I'll shoot you a DM next week and we can figure out what that looks like.

indiesquidge:
Updated to match the context exposure pattern in Tabs. No rush on the review :)

chancestrickland:
@indiesquidge Thanks! Working on this tomorrow. 👍

chancestrickland:
@indiesquidge One more thought before I merge this, curious what you think.

Combobox has `isPopoverVisible`
MenuButton has `isOpen`
Listbox has `isExpanded`

I'm thinking we could adopt `isExpanded` for all of these for consistency across the lib. Thoughts on using that name instead for this context property?

indiesquidge:
@chancestrickland `isExpanded` sounds good to me. I see that neither MenuButton nor Listbox have context hooks (yet), so that would make this PR the first public result of that property name.

Aside: not related to this PR, but I'm a bit confused by how `isExpanded` is being used in Listbox as I don't understand the difference between `current.value` and `current.context.value`:

https://github.com/reach/reach-ui/blob/d4bcfc73f27e6e618ec2eb2b0154a58bc17496f0/packages/listbox/src/index.tsx#L296-L311

chancestrickland:
`current.value` is the current state value in the state machine (see the `ListboxStates` enum in `machine.ts`, `"IDLE" | "NAVIGATING" | "SEARCHING"`, etc.). These states aren't relevant to a consumer of the component, so we just reference that state value to determine whether or not the listbox is expanded.

`current.context` is the data object in the state machine (line 321, `machine.ts`), and `value` in that object is the value of the Listbox's selected option.

The only breaking change will be in `MenuButton`, we do use `isOpen` as the key for the children render prop there.

indiesquidge:
I see now, thanks for explaining.

https://github.com/reach/reach-ui/blob/c7f7e6256561f35e82a5695ba1f5ab52766d5de0/packages/listbox/src/machine.ts#L725-L739

> The only breaking change will be in MenuButton, we do use isOpen as the key for the children render prop there.

Ah I had forgotten MenuButton exposes `isOpen` in a render prop.

Is the render prop pattern something we'd like to support in addition to the hooks moving forward? Was [that implementation][1] a one-off? Something that was implemented before we had ideas around exposing slices of internal state through hooks?

I don't know (and haven't yet spent too much time reading about) specific benefits of preferring render props over hooks, but [I'm sure there are some cases][2] where it makes sense.

[1]: https://github.com/reach/reach-ui/commit/7407794c58fac136ba67c5f009782d30c2f9020d#diff-15ead492ca6c18618236cb9abce33372R110-R114
[2]: https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-render-props

chancestrickland:
Hooks are nice, but the pattern I'd like to follow (for now) is to expose the same limited slice of context data with both. Gives users a choice, and folks may still decide to implement these in class components.

chancestrickland:
(but yes, the render prop in menu button established the pattern well before hooks were even a thing)

PR #504, opened 2020-03-09
[Combobox] Add aria-label to examples

indiesquidge:
> "If the combobox has a visible label and the combobox element is an HTML element that can be labelled using the HTML label element (e.g., the input element), it is labeled using the label element. Otherwise, if it has a visible label, the combobox element has aria-labelledby set to a value that refers to the labelling element. Otherwise, the combobox element has a label provided by aria-label." [source][1]

Given the number of ways to label the combobox, we cannot enforce or handle the aria programmatically. Outside of a runtime check, the best we can do is set good examples with proper labeling.

[1]: https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-6

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -265,7 +265,15 @@ const OptionContext = createNamedContext(
  */
 export const Combobox = forwardRefWithAs<ComboboxProps, "div">(
   function Combobox(
-    { onSelect, openOnFocus = false, children, as: Comp = "div", ...rest },
+    {
+      onSelect,
+      openOnFocus = false,
+      children,
+      as: Comp = "div",
+      "aria-label": ariaLabel,
+      "aria-labelledby": ariaLabelledby,
+      ...rest
+    },
     forwardedRef
   ) {
     let [options, setOptions] = useDescendants<HTMLElement, DescendantProps>();
@@ -319,6 +327,8 @@ export const Combobox = forwardRefWithAs<ComboboxProps, "div">(
       popoverRef,
       state,
       transition,
+      ariaLabel,
+      ariaLabelledby,
     };
 
     useEffect(() => checkStyles("combobox"), []);
@@ -360,6 +370,16 @@ export type ComboboxProps = {
    * @see Docs https://reacttraining.com/reach-ui/combobox#combobox-openonfocus
    */
   openOnFocus?: boolean;
+  /**
+   * Defines a string value that labels the current element.
+   * @see Docs https://reacttraining.com/reach-ui/combobox#accessibility
+   */
+  "aria-label"?: string;
+  /**
+   * Identifies the element (or elements) that labels the current element.
+   * @see Docs https://reacttraining.com/reach-ui/combobox#accessibility
+   */
+  "aria-labelledby"?: string;
 };
 
 if (__DEV__) {
@@ -412,6 +432,8 @@ export const ComboboxInput = forwardRefWithAs<ComboboxInputProps, "input">(
       autocompletePropRef,
       openOnFocus,
       isVisible,
+      ariaLabel,
+      ariaLabelledby,
     } = useContext(ComboboxContext);
 
     let ref = useForkedRef(inputRef, forwardedRef);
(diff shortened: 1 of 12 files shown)


chancestrickland:
Should the label not be on `ComboboxInput` rather than `Combobox` now that we've moved to support the ARIA 1.2 spec?

indiesquidge:
@chancestrickland good call, thanks for catching that. I updated the examples that had `aria-label` on the Combobox to instead have it on the ComboboxInput.

indiesquidge:
I was also curious if this type of accessibility requirement is something we should be explicitly documenting on the website since we can't handle it programmatically.

chancestrickland:
> I was also curious if this type of accessibility requirement is something we should be explicitly documenting on the website since we can't handle it programmatically.

Yes, definitely. Feel free to add some context to the documentation in this or another PR. That said, I'm kind of second-guessing my original comment. While the label should technically apply to the input, I'm thinking it might be best to use forward `aria-label` and `aria-labelledby` props from the parent. Makes it easier to keep a consistent pattern across the library where possible. If you don't get to it first I'll have this patched up and merged before the next release.

indiesquidge:
> While the label should technically apply to the input, I'm thinking it might be best to use forward `aria-label` and `aria-labelledby` props from the parent.

@chancestrickland oh I didn't realize this was a pattern we were aiming for across the library. You're saying we should have something like this?

```jsx
<Combobox aria-label="select a fruit">
  <ComboboxInput />
  <ComboboxPopover>
    <ComboboxList>
      <ComboboxOption value="Apple" />
      <ComboboxOption value="Kiwi" />
    </ComboboxList>
  </ComboboxPopover>
</Combobox>
```

Where `aria-label` (or `aria-labelledby`) is then actually applied to ComboboxInput under the hood by Reach via context?

The interesting case to consider is if they user wanted to use an actual label element.

> If the combobox has a visible label and the combobox element is an HTML element that can be labelled using the HTML `label` element (e.g., the `input` element), it is labeled using the `label` element. [source](https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-6)

Would they apply the `id` to the ComboboxInput itself still in that scenario? Is it only aria-* attributes we're looking to surface to the top-most composite component?

chancestrickland:
> Would they apply the `id` to the ComboboxInput itself still in that scenario? Is it only aria-* attributes we're looking to surface to the top-most composite component?

I think it only really makes sense for aria attributes, since there's really not a good reason for folks to use an aria label on any other children of combobox. Labeling is often the one thing we can't do for folks by default, but we still need to push for it in docs and make it as easy as possible IMO. Another benefit is that, if the aria spec changes in the future (as it did from 1.1 to 1.2) we don't have to change our API. Labels for the compound component go on the parent unless there's a good reason not to. I'm open to counter arguments here, BTW, this is just me thinking out loud!

indiesquidge:
> Labels for the compound component go on the parent unless there's a good reason not to.

@chancestrickland your arguments for this pattern make sense to me, I'm convinced.

I've [added a commit][1] that allows for `aria-label` and `aria-labelledby` to be passed in to `<Combobox>` and applied to `<ComboboxInput>` via context.

[1]: https://github.com/reach/reach-ui/pull/504/commits/f313f2dac1f481e63d340f728050f838d1c5e16b

packages/combobox/src/index.tsx
```diff
@@ -508,6 +530,8 @@ export const ComboboxInput = forwardRefWithAs<ComboboxInputProps, "input">(
         aria-controls={listboxId}
         aria-expanded={isVisible}
         aria-haspopup="listbox"
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledby}
         role="combobox"
         {...props}
```
indiesquidge: Worth pointing out that the `aria-label` and `aria-labelledby` props can still be applied directly to `<ComboboxInput>` through `props`, so this doesn't break existing uses of Combobox.

PR #514, opened 2020-03-14
[Combobox] Pass custom data of ComboboxOption to the onSelect of a Combobox

### LINKED ISSUE #502: [Combobox] Add id or custom value to ComboboxOption
## 🚀 Feature request
Upon rendering a list of items inside a `Combobox`, I'd love to pass the item id to each option without rendering the id. This would make it easier to handle the `onSelect` as I wouldn't need to filter the list by the selected item string to retrieve the id.

### Current Behavior
As far as I know, only text can be passed to the value of `ComboboxOption` which will be rendered to the DOM.

### Desired Behavior
Pass item id or custom data along.

### Suggested Solution
Add an extra `id` or `data` prop to the `ComboboxOption` and pass it along to the `onSelect` handler of the `Combobox`.

```
<Combobox onSelect={(value, data) => void alert(`You selected item with index ${data.index}!`}>
  <ComboboxInput
    aria-label="Cities"
    className="city-search-input"
    onChange={handleChange}
  />
  <ComboboxPopover className="shadow-popup">
    <ComboboxList aria-label="Cities">
      {results.slice(0, 10).map((result, index) => (
        <ComboboxOption
          key={index}
          data={{index}}
          value={`${result.city}, ${result.state}`}
        />
      ))}
    </ComboboxList>
  </ComboboxPopover>
</Combobox>
```

What do you think? If data is the second argument of `onSelect`, this wouldn't be a breaking change.

CodingDive:

This pull request:

Fixes #502

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -804,7 +804,7 @@ export const ComboboxOption = forwardRefWithAs<ComboboxOptionProps, "li">(
     }, [index, selectData, addOptionData, removeOptionData]);
 
     const handleClick = () => {
-      onSelect && onSelect(value);
+      onSelect && onSelect(value, selectData);
       transition(SELECT_WITH_CLICK, { value });
     };
 
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -26,7 +26,7 @@ import React, {
   useContext,
   useMemo,
   useReducer,
-  useState,
+  useState
 } from "react";
 import PropTypes from "prop-types";
 import {
@@ -40,13 +40,13 @@ import {
   useForkedRef,
   useUpdateEffect,
   wrapEvent,
-  noop,
+  noop
 } from "@reach/utils";
 import {
   createDescendantContext,
   DescendantProvider,
   useDescendant,
-  useDescendants,
+  useDescendants
 } from "@reach/descendants";
 import { findAll } from "highlight-words-core";
 import { useId } from "@reach/auto-id";
@@ -115,8 +115,8 @@ const stateChart: StateChart = {
         [INITIAL_CHANGE]: IDLE,
         [FOCUS]: SUGGESTING,
         [NAVIGATE]: NAVIGATING,
-        [OPEN_WITH_BUTTON]: SUGGESTING,
-      },
+        [OPEN_WITH_BUTTON]: SUGGESTING
+      }
     },
     [SUGGESTING]: {
       on: {
@@ -128,8 +128,8 @@ const stateChart: StateChart = {
         [BLUR]: IDLE,
         [SELECT_WITH_CLICK]: IDLE,
         [INTERACT]: INTERACTING,
-        [CLOSE_WITH_BUTTON]: IDLE,
-      },
+        [CLOSE_WITH_BUTTON]: IDLE
+      }
     },
     [NAVIGATING]: {
       on: {
@@ -142,8 +142,8 @@ const stateChart: StateChart = {
         [SELECT_WITH_CLICK]: IDLE,
         [SELECT_WITH_KEYBOARD]: IDLE,
         [CLOSE_WITH_BUTTON]: IDLE,
-        [INTERACT]: INTERACTING,
-      },
+        [INTERACT]: INTERACTING
+      }
     },
     [INTERACTING]: {
       on: {
@@ -154,10 +154,10 @@ const stateChart: StateChart = {
         [ESCAPE]: IDLE,
         [NAVIGATE]: NAVIGATING,
         [CLOSE_WITH_BUTTON]: IDLE,
-        [SELECT_WITH_CLICK]: IDLE,
-      },
-    },
-  },
+        [SELECT_WITH_CLICK]: IDLE
+      }
+    }
+  }
 };
 
 const reducer: Reducer = (data: StateData, event: MachineEvent) => {
@@ -168,49 +168,49 @@ const reducer: Reducer = (data: StateData, event: MachineEvent) => {
       return {
         ...nextState,
         navigationValue: null,
-        value: event.value,
+        value: event.value
       };
     case NAVIGATE:
     case OPEN_WITH_BUTTON:
       return {
         ...nextState,
-        navigationValue: findNavigationValue(nextState, event),
+        navigationValue: findNavigationValue(nextState, event)
       };
     case CLEAR:
       return {
         ...nextState,
         value: "",
-        navigationValue: null,
+        navigationValue: null
       };
     case BLUR:
     case ESCAPE:
       return {
         ...nextState,
-        navigationValue: null,
+        navigationValue: null
       };
     case SELECT_WITH_CLICK:
       return {
         ...nextState,
         value: event.value,
-        navigationValue: null,
+        navigationValue: null
       };
     case SELECT_WITH_KEYBOARD:
       return {
         ...nextState,
         value: data.navigationValue,
-        navigationValue: null,
+        navigationValue: null
       };
     case CLOSE_WITH_BUTTON:
       return {
         ...nextState,
-        navigationValue: null,
+        navigationValue: null
       };
     case INTERACT:
       return nextState;
     case FOCUS:
       return {
         ...nextState,
-        navigationValue: findNavigationValue(nextState, event),
+        navigationValue: findNavigationValue(nextState, event)
       };
 
     default:
@@ -303,7 +303,7 @@ export const Combobox = forwardRefWithAs<ComboboxProps, "div">(
       // controlling the value of ComboboxInput.
       value: "",
       // the value the user has navigated to with the keyboard
-      navigationValue: null,
+      navigationValue: null
     };
 
     const [state, data, transition] = useReducerMachine(
@@ -332,7 +332,7 @@ export const Combobox = forwardRefWithAs<ComboboxProps, "div">(
       persistSelectionRef,
       popoverRef,
       state,
-      transition,
+      transition
     };
 
     useEffect(() => checkStyles("combobox"), []);
@@ -371,7 +371,7 @@ export type ComboboxProps = {
    *
    * @see Docs https://reacttraining.com/reach-ui/combobox#combobox-onselect
    */
-  onSelect?(value: ComboboxValue): void;
+  onSelect?: (value: ComboboxValue, data?: any) => void;
   /**
    * If true, the popover opens when focus is on the text box.
    *
@@ -395,7 +395,7 @@ if (__DEV__) {
   Combobox.propTypes = {
     as: PropTypes.any,
     onSelect: PropTypes.func,
-    openOnFocus: PropTypes.bool,
+    openOnFocus: PropTypes.bool
   };
 }
 
@@ -441,7 +441,7 @@ export const ComboboxInput = forwardRefWithAs<ComboboxInputProps, "input">(
       openOnFocus,
       isExpanded,
       ariaLabel,
-      ariaLabelledby,
+      ariaLabelledby
     } = useContext(ComboboxContext);
 
     let ref = useForkedRef(inputRef, forwardedRef);
@@ -628,7 +628,7 @@ export const ComboboxPopover = forwardRef<
     // results.length or whatever).
     hidden: !isExpanded,
     tabIndex: -1,
-    children,
+    children
   };
 
   return portal ? (
@@ -738,28 +738,27 @@ if (__DEV__) {
  */
 export const ComboboxOption = forwardRefWithAs<ComboboxOptionProps, "li">(
   function ComboboxOption(
-    { as: Comp = "li", children, value, onClick, ...props },
+    { as: Comp = "li", children, value, data, onClick, ...props },
     forwardedRef: React.Ref<any>
   ) {
     const {
       onSelect,
       data: { navigationValue },
-      transition,
+      transition
     } = useContext(ComboboxContext);
-
     let ownRef = useRef<HTMLElement | null>(null);
     let ref = useForkedRef(forwardedRef, ownRef);
 
     let index = useDescendant({
       context: ComboboxDescendantContext,
       element: ownRef.current!,
-      value,
+      value
     });
 
     const isActive = navigationValue === value;
 
     const handleClick = () => {
-      onSelect && onSelect(value);
+      onSelect && onSelect(value, data);
       transition(SELECT_WITH_CLICK, { value });
     };
 
@@ -810,6 +809,13 @@ export type ComboboxOptionProps = {
    * @see Docs https://reacttraining.com/reach-ui/combobox#comboboxoption-value
    */
   value: string;
+  /**
+   * Custom data that will be passed to the `onSelect` of the `Combobox` as a
+   * second argument.
+   *
+   * @see Docs https://reacttraining.com/reach-ui/combobox#comboboxoption-data
+   */
+  data?: any;
 };
 
 if (__DEV__) {
@@ -837,14 +843,14 @@ if (__DEV__) {
 export function ComboboxOptionText() {
   const { value } = useContext(OptionContext);
   const {
-    data: { value: contextValue },
+    data: { value: contextValue }
   } = useContext(ComboboxContext);
 
   const results = useMemo(
     () =>
       findAll({
         searchWords: escapeRegexp(contextValue || "").split(/\s+/),
-        textToHighlight: value,
+        textToHighlight: value
       }),
     [contextValue, value]
   );
@@ -958,7 +964,7 @@ function useKeyDown() {
     state,
     transition,
     autocompletePropRef,
-    persistSelectionRef,
+    persistSelectionRef
   } = useContext(ComboboxContext);
 
   const { descendants: options } = useContext(ComboboxDescendantContext);
@@ -1024,7 +1030,7 @@ function useKeyDown() {
         if (state === IDLE) {
           // Opening a closed list
           transition(NAVIGATE, {
-            persistSelection: persistSelectionRef.current,
+            persistSelection: persistSelectionRef.current
           });
         } else {
           let next = getNextOption();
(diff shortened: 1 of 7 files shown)


CodingDive:
Couldn't run the tests locally and had tons of type errors in vs code. Running storybook worked fine. 
I just ran `lerna bootstrap` in root followed by `yarn test`. Should I have build before running tests?

packages/combobox/src/index.tsx
```diff
@@ -710,13 +747,15 @@ export const ComboboxOption: ComponentWithForwardedRef<
   "li",
```
CodingDive: I believe the build is failing because `ComponentWithForwardedRef` does not work correctly. You can test it by writing the following code.

`type test = ComponentWithForwardedRef<"div", { something: string }>`

TypeScript thinks the type is `any` although when I look in the `dist` folder of my project, it seems to be working fine. 🤔 

image

packages/utils/src/types.tsx
```diff
@@ -82,8 +82,15 @@ export type ComponentWithForwardedRef<
   ComponentProps
 > = React.ForwardRefExoticComponent<
   ComponentProps &
-    React.HTMLProps<React.ElementType<ElementType>> &
-    React.ComponentPropsWithRef<ElementType>
+    // Need to omit data from html props to allow a custom prop type in components.
+    // Otherwise, it always falls back to the html element attribute data: string for
+    // attaching a url of an element. However, I have never seen this used before.
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
+    Omit<
```
CodingDive: I really don't love this but the only alternative for allowing a custom "data" prop in components would be to either create a new `ComponentWithForwardedRef` type specifically for the `ComboboxOption` (right now, the component seems to be the only one using this type util anyway), or to find a different prop name than `data`.

The `data` prop of HTML elements looks like a legacy HTML prop to me. It isn't even documented on mdn (only on [w3c](https://www.w3schools.com/tags/att_object_data.asp)) and I don't think people would ever use it for developing React applications. How should we go about this?

chancestrickland:
@indiesquidge Would love to get your take on this. I haven't had a ton of time to review this PR in depth, so another set of eyes from someone familiar with our patterns and goals would be helpful! I'm planning to review it this weekend regardless @CodingDive.

packages/combobox/src/index.tsx
```diff
@@ -373,6 +376,40 @@ if (__DEV__) {
   };
 }
 
+/**
+ * Uses a ref object which stores the index as a key and custom data as value
+ * for each ComboboxOption. Hides the ref so that we can only mutate it through
+ * the returned functions. 🙈
+ */
+const useOptionDataFactory = (): {
```
indiesquidge, line 384: @CodingDive I can see the value in having the option index, but I'm curious what your thoughts on are on the use of the `data` object prop.

packages/utils/src/types.tsx
indiesquidge: I'd personally be in preference of preserving native attributes and naming props something that doesn't conflict. This also keeps us in line with the philosophy of [pretending we're HTML][1].

[1]: https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7#random-implementation-checklist

packages/combobox/src/index.tsx
```diff
@@ -782,6 +827,13 @@ export type ComboboxOptionProps = {
    * @see Docs https://reacttraining.com/reach-ui/combobox#comboboxoption-value
    */
   value: string;
+  /**
+   * Custom data that will be passed to the `onSelect` of the `Combobox` as a
+   * second argument.
+   *
+   * @see Docs https://reacttraining.com/reach-ui/combobox#comboboxoption-data
```
indiesquidge, line 834: @CodingDive looks like this anchor section is missing from the [website source page][1]. Should we update that section of the website to reflect your changes?

[1]: https://github.com/reach/reach-ui/blob/1a17f732b70a61ad57576d149312265432d8dd30/website/src/pages/combobox.mdx#comboboxoption

packages/combobox/src/index.tsx
CodingDive: Arbitrary `data` is a lot more flexible than just `index`.
I might want to pass the whole item, or just the id of an item when using a record (e.g `Record<id, Item>`).
Index is mostly relevant for arrays and even then, I wouldn't feel great to use the index of an selected item, knowing that the list we use for rendering can change at any point (filtering/sorting, etc.). As soon as the selected index is stored in state and the list is changed, the consecutive rerenders after `onSelect` might be dealing with a completely wrong or non existing item.

packages/combobox/src/index.tsx
CodingDive: Yes, will do 😊

packages/utils/src/types.tsx
CodingDive: Sure. Any ideas for a good name?

Only two I can think off at the moment that are not already taken by HTML are:

- `selectData`
- `context`

packages/combobox/src/index.tsx
indiesquidge: Apologies, I meant id rather than index. I'm still not sure I understand the full value of passing in any amount of arbitrary data (when that data could be accessed outside of the Combobox if it were keyed by an id).

Perhaps it would be good for us to show a problem that this addition would solve. Can you think of a good example case other than or in addition to the [example that alerts the selected item's index][1]?

[1]: https://github.com/reach/reach-ui/pull/514/files#diff-fcbf2f9176bec0e87f86974e979bb125R30

checkmatez:
Sorry I am late to the party, but I have an idea, borrowed from [Material-UI](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-lab/src/useAutocomplete/useAutocomplete.d.ts#L252).

How about we make Combobox generic component:
```ts
export type ComboboxProps<TValue = string> = {
  onSelect?(value: TValue): void;
  getOptionLabel?(value: TValue): string; // Function to extract presentation part to show inside ComboboxOptionText
  // ...
}
export function Combobox<TValue = string>(props: ComboboxProps<TValue>) {
  const { getOptionLabel = x => x } = props; // Default label is value itself.
  // Later we call this for each option provided, and display result in ComboboxOptionText
}
```

So in your example it could be used as:
```ts
type Option = {
   id: string;
   city: string
   state: string;
}
const Example = () => {
  return (
    <Combobox<Option>
      getOptionLabel={(option) => `${option.city}, ${option.state}`}
      onSelect={(value) => console.log(value.id)}
      // etc.
    >
       <ComboboxOption
           key={index}
            value={resultItem}
          />
    </Combobox>
  )
}
```

This way, we preserve backwards compatibility, since we have defaults for type paramater and new prop. And users are able to provide any value they want.

packages/combobox/src/index.tsx
checkmatez: I think, it just makes working with onSelect easier. You just put in some object as option value, and you receive exact same object in onSelect. No need to figure out how to first get selected id and then how to get option by id.

Check out my suggestion please - https://github.com/reach/reach-ui/pull/514#issuecomment-609458433

CodingDive:
@checkmatez I love your suggestion and like the idea of having a generic and typed component. I explored it in the beginning but gave up and settled on `any` because I couldn't figure out how to nicely pass the generic type args to child components. How would you apply TContext to the `ComboboxOption` prop? I'm not even sure if it is possible at the moment as those are completely different components. My TypeScript knowledge is limited so I don't rule it out.

I like the idea of just handling it in `onSelect`  and not needing a second argument! I don't really like the `getOptionLabel` function though. I'd probably just want to extract the label by some hard coded key like `value` and not increase the API surface area by adding another function? We could enforce TContext to have a `value: string` property when the generic type argument is provided.

checkmatez:
@CodingDive Thanks for reply. I tried to make it work for a bit, and see the problem now. In order to pass generic type down, we have to create context dynamically, in the render of Combobox.

> I'd probably just want to extract the label by some hard coded key
This will be breaking change, since type changes from string to object. If it is not a problem, I am for this change.

Looks like we have to go for the second prop.

CodingDive:
Yeah, this is exactly what I found out myself. Dynamicly created context and then the question would be how components could consume the context while somehow preserving the type definition.


> That will be a breaking change, since type changes from string to object. If it is not a problem, I am for this change.
Sorry, just to clarify despite us sticking to the second prop now where we have yet to find a fitting name, I meant a hardcoded key as long as the value is an object. The default type should have stayed a string.

packages/utils/src/types.tsx
CodingDive: Went for `selectData`

packages/combobox/src/index.tsx
CodingDive: The main use case would be to have a really big array and not have to `.find` the selected item once again. Changed the example and prop docs to reflect that other things as id or index can be passed in.

CodingDive:
This should be good to go now. I'm not sure what's going on with storybook. I couldn't try out the example locally because `yarn start` generated dozens of files and failed.

Let me know if there is anything else left to do 👍

indiesquidge:
@CodingDive thanks for updating the PR. It'd be good to get your local repo in a working state again so that you're able to run and test your changes.

In your new example, the keyboard navigation is broken.

broken-arrow-nav-combobox

This is due to the fact that you have [multiple identical values][1] in your set of options, and the internal [keydown handler uses `findIndex` to get an option's index][2], which will always return the first matching element.

Furthermore, this brings up a UX question about the example: when there are multiple options that're the same value, how would a user know which option they're selecting?

Perhaps you could share an example of the problem you're working on that this feature would help solve. From there, we might be able to create a more pragmatic example.

[1]: https://github.com/reach/reach-ui/pull/514/files?file-filters%5B%5D=.js#diff-fcbf2f9176bec0e87f86974e979bb125R43
[2]: https://github.com/reach/reach-ui/blob/073a25d18c7f923c2125f7c24336c660df0843d7/packages/combobox/src/index.tsx#L1020

CodingDive:
Thank you! I didn't know there were multiple duplicate cities 😁

In my usecase, it's often just that I ideally want to work with the item without having to find it again. One combobox in particular can have up to 20.000 entries and I'm rendering it within a virtual list. I could put them into an object and have O(1) lookup but some virtual list implementations even require you to pass an array. Therefore, in the `onSelect` callback, I just want to work with the selected item (e.g display it by putting it into state) without having to do a O(n) blocking lookup in this huge list again 😁

I understand that this is not the same problem for everybody but given the flexibility of passing arbitrary data, some may just want to work with the id to which I can change the example. 

Do you have any good idea for an example? Rendering the virtual list and thousands of items would probably be beyond the scope of an example that's easy to digest.

CodingDive:
@indiesquidge could you take another look please? 😊

indiesquidge:
@CodingDive, given our struggle to find a common example, I'm unsure why this addition should be Reach's responsibility rather than exist in user-land for the few cases that it could be necessary for.

> Therefore, in the onSelect callback, I just want to work with the selected item (e.g display it by putting it into state) without having to do a O(n) blocking lookup in this huge list again.

The O(n) lookup can be avoided by having the `optionData` be an object that we key by `ComboboxOption`'s value prop.

```diff
+ // somewhere outside of Combobox
+ const optionData = useRef({});

…

<Combobox
  onSelect={(value) => {
+   // O(1) lookup of selected option data
+   const data = optionData.current[value];
  }}
/>

…

<ComboboxList>
  {results.map((result, index) => {
    const value = `${result.city}, ${result.state}`;

+   optionData.current[value] = {
+     city: result.city,
+     state: result.state,
+     comment: "and beautiful!",
+   };

    return <ComboboxOption key={index} value={value} />;
  })}
</ComboboxList>
```

If the concern is about `optionData` growing too large (even 20,000 key-value pairs are quite manageable by [2012 browsers][1]), we can wrap `ComboboxOption` with our own component that adds/removes keys in an effect. However, if the size of a JS object is a performance bottleneck, rethinking why such a large object is needed should be considered.

> I could put them into an object and have O(1) lookup but some virtual list implementations even require you to pass an array.

We could use `Object.keys(optionData)` or `Object.values(optionData)` to convert the object to an array if that's required by other libraries.

[1]: https://joshzeigler.com/technology/web-development/how-big-is-too-big-for-json

CodingDive:
I really don't see the downside of passing arbitrary data other than maybe the fact that we're passing `any` instead of `{ id: number }`.
If we look at the linked issue https://github.com/wellyshen/use-places-autocomplete/issues/183 that integrates with the `google-map-api` I'd much rather pass `placeId: number` or `place_id: number` than `id: number` but maybe that's just me. 

Do you want me to change the PR?

CodingDive:
I strongly disagree that this is nicely handled in user land. It's not common to have an object based on value key and it seems very easy to mess this up opposed to the great DX this PR provides.

```jsx
const shoppingList = [
  {
     id: 1,
     label: 'apple',
     itemCount: 3,
  }
]

const lookupObject = shoppingList.reduce((acc, item) => ({...acc, [item.label]: item}), {});

// loop and render 80 lines below
shoppingList.map(({key, label, itemCount}) => <ComboboxOption value={label} key={key} /> )

// Might easily get refactored to

shoppingList.map(({key, label, itemCount}) => <ComboboxOption value={itemCount > 1 ? pluralize(label) : label} key={key} /> )
``` 

That's how easy it is to get bugs with your approach. Of course you can say the pluralization should've done inside the lookupObject too but one thing that stands out in this library is its great developer experience and not being opinionated about things developers can decide themselves. E.g styling. In this case it's about how I represent my data.

I think the PR is completely aligned with https://gist.github.com/ryanflorence/e5c794e6093d16a69fa88d2112a292f7

chancestrickland:
@CodingDive I totally see your point about solutions outside of Reach being potentially bug-prone, and where we can I do think we should keep things easy to work with. At the same time, we intend to keep the implementations simple where possible so that our code is less bug-prone. If a user can reasonably manage this feature outside of Reach in a wrapper component, I'd lean towards that option and then adding documentation to cover how to do it without foot-gunning yourself. So while I see the value in having arbitrary related data available in a callback in some scenarios, I'm just not sure if it's a common enough need that we need to bake it in everywhere.

chancestrickland:
I'll ask @ryanflorence to weigh in as well if he happens to see this. We chatted a bit about combobox yesterday so maybe he'll have a different perspective.

CodingDive:
Since creating the [issue](https://github.com/reach/reach-ui/issues/502) and PR, it got at least 7 thumbs up and multiple other issues linked to it. It does seem to be a pain-point of at least a few people.

Remember, there are virtually no downsides to accepting the PR. There are no breaking changes, and I didn't pull in any dependencies.

On the contrary, the upside is to no longer have to deal with `string-based` lookups (or lookups altogether depending on what data gets passed in) which can definitely be a cause for bugs.
We're improving performance (most people will stick to arrays and not use an extra hashmap/refs for the lookup) and prevent some low-hanging bugs in the process.

While I certainly can't speak for everyone, it's fixing my own problem. I'm building a design system/component-library on top of `reach-ui` and I'd really not want to miss out on the better developer experience because it's not for everyone. The same reasoning could be applied to any other optional prop or argument which has a cumbersome workaround.

To sell you over, I just created another example+test to showcase how nice it is to pass primitives. Works great with TypeScript too 😄 

`onSelect: (value: string, id: number) => void`

marcelklehr:
> given our struggle to find a common example, I'm unsure why this addition should be Reach's responsibility rather than exist in user-land for the few cases that it could be necessary for.

I think it's a very common need to separate user-facing labels from keys and ids, as the former are not always guaranteed to be unique and its just awkward to treat labels as IDs. If there are label collisions, I think, ideally, the user would even be able to provide a custom option component that renders each option, e.g. with a user profile picture. The first examples that come to my mind for label id clarity are:

 * `userId => user name`
 * `email address => name`
 * `air port code => city/airport name`
 * `country code => country name`
 * `language code => language name`
 * `URL => website title`
 * anything that you have a primary key in your database for

The only thing that I can come up with that one would directly use the label as the primary key is a tagging system (and even there you might want to translate the tags one day and *boom*, you need id's again).

CodingDive:
@marcelklehr I agree. At the moment, you can technically not have label collisions because the labels need to be unique but why not also have a unique id to be sure. Thank you for pitching in.

espebje:
Just want to chime in on this as I would love this change to get implemented. My usecase is the dropdown for autocomplete for google maps as previously linked. I just want the place_id from that box, presented with the correct labels for the users.

Having to update an internal list which replicates the one displayed in the dropbox matching to place_id is completely unnecessary and may lead to unintentional behaviour.

If the goal is to make interactions more accessible across the react universe, streamlining data handling for developers is critical or it ends up with a "meh, it passes the tests even if its not perfect behaviour", or people are like me and "meh, I'll add Reach when this PR merges".

CodingDive:
What are we waiting on?
@ryanflorence @chancestrickland can this finally get merged? 🙏

chancestrickland:
I'll defer to Ryan if he wants to take this up.

espebje:
@ryanflorence Could you give some feedback on this? It's been a week since last feedback, and the PR is soon 3 months old

chancestrickland:
Can you please open a new PR to the `develop` branch? Thanks!

CodingDive:
I don't even have confirmation if it'll get merged or not. Will open a new PR if I do..

Leo4815162342:
Any confirmation on allowing @CodingDive change to be merged? This would be a life saver!

espebje:
@chancestrickland Just dragging it along without giving any real answers on the matter does nothing but create frustration which will only sway people away from using this library, hurting what I believe to be the goal of a more accessible web.

chancestrickland:
@espebje I understand the frustration, but new features need to be vetted and tested before we release them. I have not had the time to do it yet as I'm no longer working FT on the project. Ryan hasn't either. I asked for a fresh PR so I could review thoroughly when I am able to.

Leo4815162342:
@chancestrickland thanks for the reply. @CodingDive are you still up to create new PR?

CodingDive:
Yes, will reopen it

PR #519, opened 2020-03-19
don't let tooltips grow forever

alexkrolick:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/tooltip/styles.css
+++ b/packages/tooltip/styles.css
@@ -13,4 +13,5 @@
   background: #f0f0f0;
   color: #444;
   border: solid 1px #ccc;
+  max-width: 12.5em;
 }


packages/tooltip/styles.css
```diff
@@ -13,4 +13,5 @@
   background: #f0f0f0;
   color: #444;
   border: solid 1px #ccc;
+  max-width: 12.5em;
```
alexkrolick: = 200px at browser default of 16px per em

chancestrickland:
Is there a reason this isn't better left for folks to handle in their app? We try to keep styles somewhat minimal.

alexkrolick:
> There are some basic styles to make the components usable off-the-shelf

I think most people don't expect a tooltip to flow all the way off the screen for long text. I certainly didn't, and it's not that easy to debug a transient DOM element in the devtools when it's visibility is controlled by JS and not :hover states.

chancestrickland:
I still think this makes more sense for apps to handle, as there are a number of ways you might want to deal with this and I don't think we should be opinionated about the solution. In general you should probably avoid tooltips with enough content to trigger this problem in the first place, but if you do need this solution you can add a max-width to your stylesheet or handle it with a custom positionig function with the `position` prop on `TooltipContent`.

PR #533, opened 2020-04-02
Docs: Remove obsolete `as` prop from MenuButton documentation

kohgpat:
Hello!

Thank you for the project. I've noticed the `as` prop in MenuButton documentation and found out it doesn't exist. I spend some time trying to understand what I'm doing wrong and looking for related issues, and found https://github.com/reach/reach-ui/pull/97 . So if I understand it correctly the `as` prop was postponed. In this case I would like to provide a fix, removing the mention of this prop for now.

Sincerely,
Nikolay.

### DIFF
--- a/website/src/pages/menu-button.mdx
+++ b/website/src/pages/menu-button.mdx
@@ -197,14 +197,6 @@ Any props not listed above will be spread onto the underlying button element. Yo
 </Menu>
 ```
 
-##### MenuButton `as`
-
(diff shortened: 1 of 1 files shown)


chancestrickland:
Weird, it should be there. Planning a new release this weekend so I'll double check and make sure it's added. We'll keep the docs as-is, but good catch!

kohgpat:
Got it. Thank you for the comment. I'll close the PR then.

PR #537, opened 2020-04-06
dialog: added support for the ariaHideAncestors prop.

### LINKED ISSUE #493: [dialog] Setting aria-hidden on all the things causes issues with Jaws/Chrome.
### Current Behavior

Currently, when the dialog component is used, aria-hidden="true" is set on the rest of the body elements outside the dialog. As explained by the docs, this was done so that users wouldn't be able to see the rest of the page when the dialog is open. Fortunately, this may no longer be necessary, as modern screen readers respect aria-modal="true" and role="dialog", making this technique unnecessary for most users. Unfortunately, Jaws with Chrome seems to have a hard time showing data again when aria-hidden="true" is removed from it. As a result, when a screen reader user closes the dialog, they may not see all (or perhaps any) of the content on the page.

### Expected behavior

Either don't set aria-hidden on the content outside the modal, or at least make this functionality optional.

### Your environment

| Reach Package    | Dialog        | 0.8.5        |
| React            |         | 16.12.0        |
| Browser          | Chrome        | 80        |
| Assistive tech   | Jaws        | 2020      |
| Operating System | Windows        | 10        |

cannona:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

Fixes #493

### DIFF
--- a/packages/dialog/src/index.tsx
+++ b/packages/dialog/src/index.tsx
@@ -28,6 +28,7 @@ const overlayPropTypes = {
   initialFocusRef: () => null,
   allowPinchZoom: PropTypes.bool,
   onDismiss: PropTypes.func,
+  ariaHideAncestors: PropTypes.bool,
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -92,6 +93,7 @@ const DialogInner = forwardRef<HTMLDivElement, DialogProps>(
       onDismiss = noop,
       onMouseDown,
       onKeyDown,
+      ariaHideAncestors = true,
       ...props
     },
     forwardedRef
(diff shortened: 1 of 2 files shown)


packages/dialog/src/index.tsx
```diff
@@ -92,6 +93,7 @@ const DialogInner = forwardRef<HTMLDivElement, DialogProps>(
       onDismiss = noop,
       onMouseDown,
       onKeyDown,
+      ariaHideAncestors,
```
indiesquidge, line 96: I think we'd want to default to true here as well since a [DialogOverlay can be used on its own][1]

[1]: https://reacttraining.com/reach-ui/dialog/#dialogoverlay

cannona: Thanks. Missed that.

PR #549, opened 2020-04-13
add missing type dep

hipstersmoothie:
The `warning` utils exposes the types for the `warning` package so it `@types/` should be a dependency as well.

Screen Shot 2020-04-13 at 2 27 52 PM

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -24,6 +24,7 @@
     "dist"
   ],
   "dependencies": {
+    "@types/warning": "^3.0.0",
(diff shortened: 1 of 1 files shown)


chancestrickland:
I may just go back to importing `warning` directly from the package instead. TBH there's no real reason for packages that depend on `utils` (pretty much all of them) to bring in types as a non-dev dependency.

chancestrickland:
Meh, nevermind. It's like 1 line of code. yolo

PR #570, opened 2020-04-30
[Listbox] Hide input that documentation claims is hidden

### LINKED ISSUE #569: [Listbox] "Hidden" input is visible
### Current Behavior

When I add a `name` prop to `<Listbox>` a **visible** input is rendered with the value of the input.

### Expected behavior

When I add a `name` prop to `<Listbox>` a **hidden** input is rendered with the value of the input.

### Reproducible example

Codesandbox link ["Hidden" input is visible ](https://codesandbox.io/s/reach-listbox-hidden-input-issue-m1g3g?file=/src/index.js)

### Suggested solution(s)

Change [/listbox/src/index.tsx ln 334](https://github.com/reach/reach-ui/blob/25840c3d15278e316cc74eaecf096fa01377f361/packages/listbox/src/index.tsx#L334) to `type="hidden"`?

### Your environment

| Reach Package    |    listbox     |  0.10.1       |
| React            |         |         |
| Browser          |         |         |
| Assistive tech   |         |         |
| Node             |         |         |
| npm/yarn         |         |         |
| Operating System |         |         |

ddanger:
This PR fixes #569.

**Listbox** is not hiding the generated `input`. But documentation and tests indicate the rendered `input` is hidden. 

- This PR hides it by making it `type="hidden"`.
- Unit tests have been updated to ensure the input is hidden.
- No storybook was added because this is a bug fix. The change does hide the input that used to be visible in the `With a Form` storybook.

---

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/listbox/__tests__/listbox.test.tsx
+++ b/packages/listbox/__tests__/listbox.test.tsx
@@ -197,6 +197,10 @@ describe("<Listbox />", () => {
           </Listbox>
         );
         expect(container.querySelector("input")).toBeTruthy();
+        expect(container.querySelector("input")).toHaveAttribute(
+          "type",
+          "hidden"
+        );
       });
     });
 
@@ -218,7 +222,12 @@ describe("<Listbox />", () => {
             </Listbox>
           </div>
         );
-        expect(container.querySelector("input")).toBeTruthy();
+        // Be careful not to match the `input[type=text]` added in this test
+        expect(container.querySelector("input[type=hidden]")).toBeTruthy();
+        expect(container.querySelector("input[type=hidden]")).toHaveAttribute(
+          "form",
+          "my-form"
+        );
       });
     });
 
@@ -232,6 +241,10 @@ describe("<Listbox />", () => {
           </Listbox>
         );
         expect(container.querySelector("input")).toBeTruthy();
+        expect(container.querySelector("input")).toHaveAttribute(
+          "type",
+          "hidden"
+        );
         expect(container.querySelector("input")).toHaveAttribute("required");
       });
     });
--- a/packages/listbox/__tests__/listbox.test.tsx
+++ b/packages/listbox/__tests__/listbox.test.tsx
@@ -196,11 +196,7 @@ describe("<Listbox />", () => {
             <ListboxOption value="lengua">Lengua</ListboxOption>
           </Listbox>
         );
-        expect(container.querySelector("input")).toBeTruthy();
-        expect(container.querySelector("input")).toHaveAttribute(
-          "type",
-          "hidden"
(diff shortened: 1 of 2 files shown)


packages/listbox/__tests__/listbox.test.tsx
```diff
@@ -197,6 +197,10 @@ describe("<Listbox />", () => {
           </Listbox>
         );
         expect(container.querySelector("input")).toBeTruthy();
+        expect(container.querySelector("input")).toHaveAttribute(
```
chancestrickland, line 200: Can we instead test that the input is visibly hidden with `expect(container.querySelector("input")).not.toBeVisible()`? I'll probably change the implementation of this in the near future to address arrow form navigation on iOS.

packages/listbox/__tests__/listbox.test.tsx
```diff
@@ -218,7 +222,12 @@ describe("<Listbox />", () => {
             </Listbox>
           </div>
         );
-        expect(container.querySelector("input")).toBeTruthy();
+        // Be careful not to match the `input[type=text]` added in this test
+        expect(container.querySelector("input[type=hidden]")).toBeTruthy();
+        expect(container.querySelector("input[type=hidden]")).toHaveAttribute(
```
chancestrickland, line 227: Let's remove these assertions, see comment above about future implementation changes.

chancestrickland:
Thank you! Just a few small testing changes and I'll get this merged 👍

packages/listbox/__tests__/listbox.test.tsx
ddanger: Sure. Do you mean restoring the test to how it was? Or just removing all assertions in this test?

chancestrickland: Yeah, we can revert these lines to the one line from before. Let's just keep the check for a standard input and use the `not.toBeVisible` method to make sure it's hidden.

ddanger: There are 2 inputs, the first added by the test (`<input type="text" name="name" />` not hidden) and the other added by `listbox` (which is hidden).

So the test you propose fails.

Do you like this option which will work to get that second input?
```
 expect(container.querySelectorAll("input")[1]).not.toBeVisible();

```

ddanger: I went ahead and pushed that up along with fixing the other line.

ddanger:
Let me know if you're waiting on something from me. I think I've responded to your request for changes.

Issue #165, opened 2019-05-06
Tab buttons need `type='button'`


flybayer:
I'm launching the Dialog from inside a form. The default tab buttons have the default, implicit type of "submit" so clicking a tab submits the form.

Workaround is manually set `type='button'` on each `<Tab>`

olistic:
I just run into this issue. Happy to submit a PR for this, just lemme know!

chancestrickland:
Fixed in https://github.com/reach/reach-ui/pull/304

1000hz:
this seems to have regressed in https://github.com/reach/reach-ui/commit/a2ab2be8b0ca122e3fa03a00e24d78b757b48e35

chancestrickland:
Per my note in the linked PR, this was removed because our type check isn't reliable. It might still be worth it to have this as a default, but if a custom component is passed to `as` then the end user will still have to manually set `type="button"` if a button element is ultimately created.

PR #572, opened 2020-05-03
Use type button by default in Tab component

rubenmoya:
This PR fixes https://github.com/reach/reach-ui/issues/165, it was first fixed in https://github.com/reach/reach-ui/commit/5f75dd525385fc2d800205ff99dae205a1bce6b4 but then broken in https://github.com/reach/reach-ui/commit/a2ab2be8b0ca122e3fa03a00e24d78b757b48e35.

I've added a couple of tests but I'm not sure if they're what you expect so feedback is welcome!
_________

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/tabs/src/index.tsx
+++ b/packages/tabs/src/index.tsx
@@ -455,6 +455,8 @@ export const Tab = forwardRefWithAs<
     context: TabsDescendantsContext,
     disabled: !!disabled,
   });
+  const htmlType =
+    Comp === "button" && props.type == null ? "button" : props.type;
 
   const isSelected = index === selectedIndex;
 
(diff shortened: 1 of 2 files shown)


packages/tabs/src/index.tsx
```diff
@@ -455,6 +455,8 @@ export const Tab = forwardRefWithAs<
     context: TabsDescendantsContext,
     disabled: !!disabled,
   });
+  const htmlType =
+    Comp === "button" && props.type == null ? "button" : props.type;
```
chancestrickland, line 459: The reason this was removed to begin with is because this check isn't reliable. `Comp` might be a styled component, or a wrapper component of some kind, in which case we'd lose that default setting. Of course if we pass the attribute along to a non-button element then we have invalid HTML. May still be worth it to have either way.

rubenmoya: Maybe we could keep the check, since it would cover at least some cases, and add some information in the docs about the edge cases?

PR #609, opened 2020-06-06
[popover] Fix stuck focus

michaeldfoley:
When the `targetRef` is the last tabbable element outside the popover, `getElementAfterTrigger` returns the first tabbable element inside the portal causing the focus to stay trapped inside the popover. What we want is for `getElementAfterTrigger` to return an element only if it is not inside the popover.

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/popover/src/index.tsx
+++ b/packages/popover/src/index.tsx
@@ -241,7 +241,11 @@ function useSimulateTabNavigationForReactTree<
       elements && triggerRef.current
         ? elements.indexOf(triggerRef.current)
         : -1;
-    return elements && elements[targetIndex + 1];
+    const elementAfterTrigger = elements && elements[targetIndex + 1];
+    return popoverRef.current &&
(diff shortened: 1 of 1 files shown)


chancestrickland:
@michaeldfoley Can you reproduce a test case using the CodeSandbox link generated above? I'm happy to merge this if I can get a better understanding of the problem with a simple example.

michaeldfoley:
In [this example](https://codesandbox.io/s/optimistic-dirac-wslon?file=/src/index.js) the focus gets stuck in the popover when there are no tabbable items after the targetRef. 

Here is the [same example with my PR applied.](https://codesandbox.io/s/quirky-wave-t2s26?file=/src/index.js) After the popup, the focus moves into the chrome as expected.

chancestrickland:
Looks good to me, thank you!

PR #620, opened 2020-06-17
Add prop to dialog to bypass focus lock

### LINKED ISSUE #615: Dialog's focus lock prevents iframe from having focus
## 🐛 Bug report

Imagine a focus-locked area of a web page (in this case a reach-ui dialog). Then we open an iframe that's actually mounted outside of the focus-locked area, though visually it appears as a dialog itself (though this iframe is not using `@reach/dialog`, the part that is focus-locked is what is really a `@reach/dialog` in this example).

See codesandbox example linked to below.

### Current Behavior

The focus-locked `@reach/dialog` instance is too aggressive in keeping the focus for itself, preventing the iframe from having it.

### Expected behavior

I understand the current behavior. But I'd like that `@reach/dialog` would allow me to tell it otherwise.

### Reproducible example

[CodeSandbox Template](https://codesandbox.io/s/late-dream-ywt3o?file=/src/index.js)

Try to type inside the Wikipedia search field that opens inside the iframe.

### Suggested solution(s)

Maybe by exposing the whilelist prop of the react focus lock somehow.

### Additional context

Related to #536. That one fixed the case where the focus-locked area is inside an iframe, and was preventing from accessing things outside the iframe. However, this case is when the focus locked area is outside the iframe, and prevents focus from being set to things inside an iframe.

### Your environment

| Reach Package    |  dialog |  0.10.3 |
| React            |         | 16.13.1 |
| Browser          | Chrome, Firefox | latest |
| Assistive tech   |         |         |
| Node             |         |         |
| npm/yarn         |         |         |
| Operating System |         |         |

gnapse:
Closes #615 

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/dialog/examples/custom-focus-lock.example.tsx
+++ b/packages/dialog/examples/custom-focus-lock.example.tsx
@@ -1,0 +1,133 @@
+import React from "react";
+import { DialogOverlay, DialogContent } from "@reach/dialog";
+import "@reach/dialog/styles.css";
+import FocusLock from "react-focus-lock";
+import Portal from "@reach/portal";
+
+let name = "With customized Focus Lock";
+
+const FRAME_OVERLAY_STYLES: React.CSSProperties = {
+  position: "fixed",
+  bottom: "0px",
+  top: "0px",
+  right: "0px",
+  left: "0px",
+  zIndex: 100000,
+  backgroundColor: "rgba(160, 160, 160, 0.2)",
+};
+
+const FRAME_STYLES: React.CSSProperties = {
+  display: "block",
+  backgroundColor: "white",
+  border: "1px solid rgb(172, 172, 172)",
+  width: "600px",
+  height: "400px",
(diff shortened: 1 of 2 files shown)


gnapse:
@chancestrickland can you share some details on when would be reasonable to expect this to be released? I wonder if I can use the dependency directly from github in the mean time.

chancestrickland:
Next day or so! Needed to add some docs and make some slight adjustments after I wrote them up (we won't pass this prop to the high-level `Dialog` because you need to compose the parts to replace the focus lock).

chancestrickland:
@gnapse If you absolutely need it now, you can use `patch-package` in the mean time, then update as soon as this is cut.

chancestrickland:
@gnapse Take a peek at 0.10.5, let me know if you have any issues.

gnapse:
@chancestrickland it works like a charm. Thanks you!

PR #628, opened 2020-06-27
Combobox custom data

### LINKED ISSUE #502: [Combobox] Add id or custom value to ComboboxOption

CodingDive:

This pull request:

Fixes #502 

Reopened after https://github.com/reach/reach-ui/pull/514 was closed to rename branches


Leo4815162342:
Summoning @chancestrickland as well

tonypee:
Its a shame that this hasn't made it through, its a core requirement of a comboxbox... guess Im off to react-select

indiesquidge:
@tonypee can you please explain your use-case, preferably using the [CodeSandBox template][1]? Have you tried some of the suggested [alternative solutions][2], and if so, could you explain why they don't work for your case?

[1]: https://codesandbox.io/s/reach-ui-template-xonfs
[2]: https://github.com/reach/reach-ui/pull/514#issuecomment-630504047

tonypee:
@indiesquidge Im trying to create an autocomplete selectbox for a form which can have existing data (editing the item). So i need to be able to set the value of the selected item on load. Then be able to change that. 
It seems weird that the onSelect is giving me the 'label' value basically which i guess i can match against the options to find the value. 
Maybe im not understanding things, but it seems that the 'value' on <ComboboxOption> is actualy the label, and 'key' is the value, but it isnt what is returned in the onChange or onSelect, whats returned is the label. So i guess if i have a an actual 'value' that i assign, i need to first process that to be the same as a label... based on matching it against my options (a value/label array)

tonypee:
Basically i want to make it a controlled component

igloude:
Agreed @tonypee, this is so close to being exactly what we need.

igloude:
Any specific reason we can't reopen this one @indiesquidge?

indiesquidge:
@igloude @tonypee, can you please explain your use-case using a [CodeSandbox][1]? Have you tried some of the suggested [alternative solutions][2], and if so, could you explain why they don't work for your case?

[1]: https://codesandbox.io/s/reach-ui-template-xonfs
[2]: https://github.com/reach/reach-ui/pull/514#issuecomment-630504047

igloude:
@indiesquidge, thanks for the reply! Here's [a stripped-down version of our use-case](https://codesandbox.io/s/reach-ui-628-use-case-476kd?file=/src/index.js).

We're in a place where we need a unique value for each option - obviously - but we want to display associated data that may not be unique as the input's value after an option has been selected. There are 2 scenarios in the sandbox, toggle the comments on `setValue()` inside `handleSelect` to see what our current options are using Reach's combobox - neither of which work for us.

indiesquidge:
@igloude thanks for the example!

>  We want to display associated data that may not be unique as the input's value after an option has been selected.

In your example the associated data is still unique … is there a reason you cannot use ```<ComboboxOption value={`Not exactly ${item.id}\`}>```?

If your goal is to programmatically change the input value to something other than the selected option value, I'm still not sure how this issue would be fixed by custom option data since you'd still be setting the input value to something other than an option upon selection.

Alternatively, is there a reason it's the combobox input value that needs to be changed to something other than an option value or user input? That's a bit of an unexpected pattern for a combobox given that selecting one of the suggestions or typing a value are the only two behaviors of this widget as defined by the [WAI ARIA practices][2] this component adheres to. Perhaps the associated data could be held in some other state?

```js
const [associatedData, setAssociatedData] = React.useState()

const handleSelect = v => {
  // if you organized your data as an object keyed by id, 
  // you could get constant lookup rather than using find
  const associatedData = data.find(datum => datum.id === v)
  setAssociatedData(associatedData)
  setValue(v)
}
``` 

Taking a further step back, your options don't include the actual value in them

```js
<ComboboxOption value={item.id} key={item.id}>
  {item.title} - {item.group}
</ComboboxOption>
```

Reach recommends to have the value at least _somewhere_ in the option

> But if you need to control a bit more, you can put whatever children you want, but make sure to render a ComboboxOptionText as well, so the value is still displayed with the text highlighting on the matched portions. [[source][1]]

Otherwise, how does a user—typing into the input or navigating with the arrow keys—know which option they are selecting without browsing the entire list?

[1]: https://reach.tech/combobox/#comboboxoption-children
[2]: https://www.w3.org/TR/wai-aria-practices-1.2/#combobox

igloude:
Hey @indiesquidge, appreciate the quick response!

> In your example the associated data is still unique … is there a reason you cannot use <ComboboxOption value={`Not exactly ${item.id}\`}>?

Yes, for design reasons we'd like to not have the ID in the option text, but it needs to be the submitted value. So either we would like the component to be able to:

1. Have 2 sets of "values" per option - one that is visually selected, and one that is functionally submitted.
or
2. onSelect, submit the option value and then alter the input value to the desired string for visual purposes, and also keep the popover closed until further user interaction.

> Alternatively, is there a reason it's the combobox input value that needs to be changed to something other than an option value or user input?

I think this assumption is my fault in the example given. Consider this object:
```
{
      "id": 1234,
      "title": "Dogs are the best!",
      "group": "Pet Lovers"
}
```

In our case, the ID of `1234` means absolutely nothing to the user, but our API requires it to be submitted in the form. Also, the `title` and `group` here cannot be guaranteed to be unique. We're looking for the Combobox to be able to submit the id of `1234`, but display `Dogs are the best!` as the input's "value". Does that clarify our use case a bit more?

Thanks again for the conversation!

indiesquidge:
@igloude, thanks for the further explanation and also for the cordial conversation :)

From a design perspective, it makes sense why you don't wish to have the id as part of the option text, but I'm a bit lost on this part

> Also, the title and group here cannot be guaranteed to be unique.

If you aren't including the id as part of the option value and are instead using the title and/or group, but the title and group are not guaranteed to be unique, how are you going to avoid visual duplicates in your dropdown list? How would the user know which option they are selecting?

> We're looking for the Combobox to be able to submit the id of 1234, but display Dogs are the best! as the input's "value".

This is why I recommended keeping track of the [controlled input value separate from the selection data][1] and potentially using your data source to [map input values to the data you'd like to submit][2].

[1]: https://github.com/reach/reach-ui/blob/a1ed8e8a55cd4c76292cb9a5fe5697930cc3f6e7/packages/combobox/examples/controlled.example.js
[2]: https://github.com/reach/reach-ui/pull/514#issuecomment-630504047

chaance:
I'm finally getting a bit of free time to look at this. Thank you all for your patience.

@indiesquidge The way I understand it is that the original proposal is for a way to pass arbitrary data for a particular option into `onSelect` independent of its display value. This makes sense to me because, while the `value` should absolutely still be a unique identifier and visible to the consumer, it may still be independent of the arbitrary data they need to access in the callback. Of course you can do this if you have a lookup table of some kind, but if the visible value is different from the identifier you want to access in `onSelect` you might have to do a second (potentially expensive) search through the collection to find it, even though you already found it in the sorting process.

@CodingDive did I summarize this correctly?

As an aside: @igloude I would take another look at your design decisions if you have multiple options that aren't unique. As @indiesquidge mentioned this sounds like this could create a problematic UX. You may have an elegant solution already, I just wanted to echo that point as values for options should be uniquely identifiable.

chaance:
I may need to re-review https://github.com/reach/reach-ui/pull/514#issuecomment-630504047 as there is a lot of discussion there I wasn't able to keep up with before. Apologies if you've already answered my questions above!

indiesquidge:
@chaance thanks for chiming in. I'm okay if we want to support this, but that lookup doesn't need to be expensive, and can't all of the desired behavior can be done in user code with JS data massaging, as outlined [here](https://github.com/reach/reach-ui/pull/514#issuecomment-630504047)?

```js
const someGiantCityCollection = [
  {
    id: 1000000,
    city: "Abbeville",
    state: "Louisiana"
  },
  {
    id: 1000001,
    city: "Aberdeen",
    state: "Maryland"
  },
];

const resultsByValue = someGiantCityCollection.reduce((acc, location) => {
  const value = `${location.city}, ${location.state}`;
  acc[value] = location;
  return acc;
}, {});

<Combobox onSelect={(value) => {
  const id = resultsByValue[value].id;
  someExternalFunctionThatNeedsTheId(id)
}}>
```

chaance:
> I'm okay if we want to support this as an API addition nicety, but that lookup doesn't need to be expensive even for a giant list of data, and can't all of the desired behavior be done in user code with JS data massaging ~as outlined [here](https://github.com/reach/reach-ui/pull/514#issuecomment-630504047)~ (oops I see you reference this in your second comment)?

This is kind of where my head is at currently, I just want to keep this open to hear the use cases and make sure I have a clear understanding of the problem. It seems like something a lot of people would like.

CodingDive:
> @CodingDive did I summarize this correctly?

Yes.

> @CodingDive am I missing any important cases here where that search is causing a problem, or do you see this purely as an API improvement?

Yes, the workaround of mapping values to arbitrary data in an object works. Nonetheless, I believe it's an API improvement that offloads code from users if they want to build on top of reach-ui and improves developer ergonomics. Almost the exact same discussion was already had in the former pull request (also see my comment of how this API will inevitably reduce the potential for bugs https://github.com/reach/reach-ui/pull/514#issuecomment-631309913).


> I'm okay if we want to support this as an API addition nicety

Then get a hold of Ryan or merge the PR already! 😄

chaance:
> Yes, the workaround of mapping values to arbitrary data in an object works. Nonetheless, I believe it's an API improvement that offloads code from users if they want to build on top of reach-ui and improves developer ergonomics. Almost the exact same discussion was already had in the former pull request (also see my comment of how this API will inevitably reduce the potential for bugs [#514 (comment)](https://github.com/reach/reach-ui/pull/514#issuecomment-631309913)).

I appreciate this context and caught up with the previous discussion, thanks for that! I spent a good deal of time considering the arguments and reviewing the code, and this is where I think I've landed.

1. We consider public APIs very carefully. Additional API surface area adds to the internal complexity and maintenance overhead. That doesn't disqualify anything that adds to the API, it just means that, as a rule, we're going to be pretty nitpicky and careful about adding API features, especially when they could otherwise be created by consumers.

2. In this case, the API you want can actually be created exactly as you've designed by wrapping the existing combobox components with no internal changes. I copied most of your code here and adapted it slightly to work with the current API. We don't really need to expose the index here because values should always be unique anyway. https://codesandbox.io/s/reach-ui-combobox-custom-onselect-data-6vjcq?file=/src/App.tsx (if you see any reason this slight alteration wouldn't work for your use case, let me know, but we do generally expect that all values should be unique)

> Then get a hold of Ryan or merge the PR already! 😄

3. I did ping Ryan, but his focus has been on Remix for a while and he's delegated Reach to me for now. This project is still his baby, so if he decides to look at this and has a different take, he's welcome to overrule and merge away. In the mean time, I still think this functionality still perfectly achievable with existing APIs. I'm working on it for free in my spare time at the moment, so my priority is fixing bugs and general maintenance over new features.

PR #640, opened 2020-07-06
[Popover] Don't continuously observeRect on Popovers that are hidden

### LINKED ISSUE #636: MenuButton performance issues
### Current Behavior
It seems that simply having a MenuButton on the page (not expanded) results in a `getBoundingClientRect` call on every frame (see screenshot). On other pages (e.g. reach-alert), the profile on the right side is nice and empty.
image

### Expected behavior
If the menu isn't open, there is no reason to be observing and measuring it. (It would be nice if it wasn't re-measured even if it _was_ open)

### Reproducible example
[Reach's own docs](https://reacttraining.com/reach-ui/menu-button)

### Suggested solution(s)
I'm not too with what's under the hood, but perhaps add a prop to the popover component that allows enabling / disabling measurement, and control this prop based on menu state?

### Additional context
We ran into this issue while profiling our own application. Even when the window isn't in focus, the idle cpu usage for this chrome tab sits at 10% due to this issue.

### Your environment
| Reach Package    |    menu-button     |    0.10.5     |
| Browser          |    Chrome     |       83    |

cpmsmith:
This pull request:

Fixes #636.

Since, in order to have an `aria-controls` attribute,  `MenuButton` merely hides its `Popover` rather than not rendering it, the `Popover` is constantly running `getBoundingClientRect` by way of `useRect` whenever there's a `MenuButton` on the page. This change makes `Popover` pass the `observe` parameter to `useRect` according to its `hidden` prop, and adds a storybook example for demoing it.

I added the storybook in order to try and break it, i.e. by changing the location of the target while `observe` was false, but couldn't manage to do so.

### DIFF
--- a/packages/popover/examples/hidden.example.js
+++ b/packages/popover/examples/hidden.example.js
@@ -1,0 +1,54 @@
+import React, { useRef, useState, useCallback } from "react";
+import Popover from "@reach/popover";
+
+let name = "Hidden";
+
+function Example() {
+  const ref = useRef(null);
+  const [open, setOpen] = useState(false);
+  const updateOpen = useCallback(
+    (e) => {
+      setOpen(e.target.checked);
+    },
+    [setOpen]
+  );
+
+  return (
+    <div>
+      <textarea placeholder="resize me to move stuff around" />
+      <label style={{ display: "inline-block" }}>
+        <input ref={ref} type="checkbox" onChange={updateOpen} />
+        Open popover
+      </label>
+      <Popover hidden={!open} targetRef={ref}>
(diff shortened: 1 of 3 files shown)


cpmsmith:
Here's a trace of the basic `MenuButton` storybook example before the change:

image

And after, with the menu open from ~2.7–5.0s:

image

sciyoshi:
@chancestrickland @ryanflorence anything left before this PR can be merged? We've been running into this and it causes some performance issues - having a way to hide the popovers when not rendered would be great.

PR #650, opened 2020-08-03
@reach/tabs: Fix wrap of blur event in Tab component

lucasecdb:
This pull request:

- [ ] Creates a new package
- [ ] Other


packages/tabs/__tests__/tabs.test.tsx
```diff
@@ -531,5 +531,43 @@ describe("<Tabs />", () => {
       // fireEvent.keyDown(tabList, { key: "ArrowRight", code: 39 });
       // expect(getByText("Tab 2")).toHaveFocus();
     });
+
+    it("correctly calls focus and blur events on Tab component", () => {
```
lucasecdb, line 535: I added this test but I don't know if it is necessary for these changes, because I haven't seen similar tests to this one in this test file

Issue #638, opened 2020-07-03
CI builds are slow


mjackson:
### Current Behavior

`yarn build` currently takes about 9 minutes to complete in CI. See e.g. https://travis-ci.com/github/reach/reach-ui/jobs/357412345

### Expected behavior

I'd expect to be able to run the tests in under a minute.

### Reproducible example

N/A

### Suggested solution(s)

I'm not sure what could be slowing it down, but I've heard from some folks that ts-node is kinda slow. Are we invoking it to build each individual test?

chancestrickland:
Agreed, it's kind of painful. I'm using Github actions in another repo and it seems much faster, but I haven't investigated it too deeply. I think creating a fork and removing TS from the build scripts altogether would be a good test though.

chancestrickland:
BTW @mjackson I did manage to get the build time reduced a bit by using Node directly, but it's not really significant. Any insight from your work on the new React Router v6 build scripts?

mjackson:
One of the main things I did in the router repo to speed up testing and builds was to ditch `lerna workspace run` (or whatever it's called, can't remember). All the build + test infrastructure is able to work with the monorepo format, so e.g. I only have a single instance of Jest/Rollup running when I need to test/build. This eliminates startup time for spinning up individual processes, which can be significant in a repo with many packages.

PR #654, opened 2020-08-20
Speed up build script

chancestrickland:
Addresses #638. Builds a great deal faster locally by running a single node process vs. separate processes via `lerna run`.


chancestrickland:
Comparing the most recent CI build in Travis (only looking at the build script here, not overall CI build time):

**Before this PR:** 314.11s - https://travis-ci.com/github/reach/reach-ui/builds/174478011
**After:** 163.68s - https://travis-ci.com/github/reach/reach-ui/builds/180612849

PR #662, opened 2020-08-31
@reach/tooltip: Export `positionTooltip` function

travigd:
This allows users to re-use this logic in their own apps.

In particular, I had a use case where I wanted to change the tooltip offset and the only way I could do so was to copy the definition of the `positionDefault` function.

**Note:** I didn't test this (this PR was made from the web interface), just curious if this is something that would be acceptable. Alternatively (and/or additionally), we could add an `offset` prop to the `Tooltip` component.

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/tooltip/src/index.tsx
+++ b/packages/tooltip/src/index.tsx
@@ -490,7 +490,7 @@ const TooltipContent = forwardRefWithAs<TooltipContentProps, "div">(
       id,
       isVisible,
       label,
-      position = positionDefault,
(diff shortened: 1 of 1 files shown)


travigd:
Update: Would also be nice to export `PRect` type.

travigd:
Can this be re-opened? @chaance

chaance:
Sorry for the delay here, but this looks good to me. Thanks!

PR #671, opened 2020-09-23
Reduce slider padding

danny-andrews:
The default styling adds a ton of extraneous spacing to the slider, which is especially pronounced when adding a label (which one should always do).

Before:
image

After:
image

With 22px of top and bottom margin, the touch target is still 44px, which is in line with UX and accessibility guidelines, but without the extra padding, which most users of this component are going to have to override.

This pull request:

- [ ] Creates a new package
- [X] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/slider/styles.css
+++ b/packages/slider/styles.css
@@ -6,8 +6,8 @@
   width: 350px;
   max-width: 100%;
   height: 4px;
-  padding-top: 30px;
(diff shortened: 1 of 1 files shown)


danny-andrews:
Sorry, used wrong branch.

PR #674, opened 2020-09-24
[disclosure] Export DisclosurePanelProps type + context hook

CPatchane:
Hi here 👋 

I guess this was a typo that that type wasn't exported like for other components props type.


chaance:
Thank you for the PR! I made a number of unrelated changes separate to this, but this was done in https://github.com/reach/reach-ui/commit/1a8c4c237f54220b227aa773b7775762d91941db which is now merged. You should see it in the next release. Sorry for the overlap here!

PR #683, opened 2020-10-06
Stop Combobox Esc propagation

### LINKED ISSUE #395: Escape key closes multiple components
### Current Behavior

When you have an open combobox inside of an open dialog and you hit the escape key, it closes both the combobox and the dialog.

Screenshot 2019-11-29 at 17 14 30

### Expected behavior

Instead of closing both the combobox and the dialog, it should only close the combobox when hitting the escape key.

### Reproducible example

https://codesandbox.io/embed/reach-ui-template-7uyon

### Suggested solution(s)

Maybe use a global stack that keeps track of all dismissible UI elements that listen to "escape". Every time you hit "escape" it could call `pop()` to only call the `onDismiss` of the "deepest" UI element.

### Additional context

Perhaps the same issue occurs with other combinations of UI elements.

### Your environment

| Software                 | Name/Version(s) |
| ------------------------ | --------------- |
| @reach/combobox | 0.1.1 |
| @reach/dialog | 0.2.9 |
| React | 16.10.1 |
| Browser | Google Chrome 78.0.3904.108|
| Assistive tech | none |
| Node | 11.15.0 |
| npm/yarn | Yarn 1.17.3 |
| Operating System | MacOS Catalina 10.15.1 |

frontsideair:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

-------

This PR is to fix #395 but seems like the suggested fix doesn't seem like it works. At least the test would be useful.

### DIFF
--- a/packages/combobox/__tests__/combobox.test.tsx
+++ b/packages/combobox/__tests__/combobox.test.tsx
@@ -11,6 +11,7 @@ import {
   ComboboxInputProps,
   useComboboxContext,
 } from "@reach/combobox";
+import { Dialog } from "@reach/dialog";
 import matchSorter from "match-sorter";
 import cities from "../examples/cities";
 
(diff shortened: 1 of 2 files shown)


packages/combobox/__tests__/combobox.test.tsx
```diff
@@ -219,9 +220,36 @@ describe("<Combobox />", () => {
...
+      let { getByTestId, queryByTestId } = render(<BasicDialog />);
+      let input = getByTestId("input");
+
+      await userEvent.type(input, "e");
+
+      expect(getByTestId("list")).toBeInTheDocument();
+
+      await userEvent.type(input, "{esc}");
+
+      expect(queryByTestId("list")).not.toBeInTheDocument();
```
indiesquidge, line 235: @frontsideair thanks for your PR contribution!

The issue with your test seems to be the use of `queryByTestId`. Using `ByTestId` is [not recommended][1] if there are other queries more inline with how the software is used by people or screen readers. We can use `ByRole` for most things associated with Combobox, and indeed your tests pass with these query selectors.

Given #395 also cares about the event propagating, we should also assert that our dialog remains open when escape is pressed. Perhaps something like this:

```js
it("should not close the dialog when Esc key is pressed", async () => {
  let { getByRole, queryByRole } = render(<BasicDialog />);
  let input = getByRole("combobox");

  expect(getByRole("dialog")).toBeInTheDocument();

  await userEvent.type(input, "e");

  expect(getByRole("listbox")).toBeInTheDocument();

  await userEvent.type(input, "{esc}");

  // escape should close our listbox
  expect(queryByRole("listbox")).not.toBeInTheDocument();

  // but our dialog should still be open
  expect(queryByRole("dialog")).toBeInTheDocument();
});
```

[1]: https://testing-library.com/docs/dom-testing-library/api-queries#bytestid

PR #692, opened 2020-10-28
[Menu Button] Add extra prop for Menu component to avoid the page scroll on menu focus

thiniulian:

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/menu-button/examples/avoid-page-scroll-on-focus.example.js
+++ b/packages/menu-button/examples/avoid-page-scroll-on-focus.example.js
@@ -1,0 +1,36 @@
+import React from "react";
+import { Menu, MenuList, MenuButton, MenuLink } from "@reach/menu-button";
+import "@reach/menu-button/styles.css";
+
+let name = "Avoid Page Scroll on Menu Focus";
+
+function Example() {
+  const menus = [1, 2, 3, 4];
+  const links = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+  return (
+    <div>
+      <h2>Home</h2>
+      {menus.map((mainIndex) => (
+        <Menu key={`${mainIndex}`} keepScrollPosition={true}>
+          <MenuButton>
+            Actions {mainIndex} <span aria-hidden="true">▾</span>
+          </MenuButton>
+          <MenuList>
+            {links.map((secondIndex) => (
+              <MenuLink
+                href={`#/${mainIndex}${secondIndex}`}
+                key={`${mainIndex}_${secondIndex}`}
+              >
+                Link {secondIndex}
+              </MenuLink>
+            ))}
+          </MenuList>
+        </Menu>
+      ))}
+    </div>
+  );
+}
+
+Example.story = { name };
+export const Comp = Example;
+export default { title: "MenuButton" };
--- a/packages/menu-button/src/index.tsx
+++ b/packages/menu-button/src/index.tsx
@@ -96,7 +96,11 @@ const initialState: MenuButtonState = {
  *
  * @see Docs https://reach.tech/menu-button#menu
  */
-export const Menu: React.FC<MenuProps> = ({ id, children }) => {
+export const Menu: React.FC<MenuProps> = ({
+  id,
+  children,
+  keepScrollPosition = false,
+}) => {
   let buttonRef = useRef(null);
   let menuRef = useRef(null);
   let popoverRef = useRef(null);
@@ -145,7 +149,13 @@ export const Menu: React.FC<MenuProps> = ({ id, children }) => {
       // @ts-ignore
       window.__REACH_DISABLE_TOOLTIPS = true;
       window.requestAnimationFrame(() => {
-        focus(menuRef.current);
+        if (keepScrollPosition) {
+          const scrollOptions = { left: window.scrollX, top: window.scrollY };
+          focus(menuRef.current);
+          window.scrollTo(scrollOptions);
+        } else {
+          focus(menuRef.current);
+        }
       });
     } else {
       // We want to ignore the immediate focus of a tooltip so it doesn't pop
@@ -195,6 +205,7 @@ export interface MenuProps {
         }
       ) => React.ReactNode);
   id?: string;
+  keepScrollPosition?: boolean;
 }
 
 if (__DEV__) {
--- a/website/src/pages/menu-button.mdx
+++ b/website/src/pages/menu-button.mdx
@@ -83,9 +83,10 @@ The wrapper component for the other components. No DOM element is rendered.
 
 #### Menu Props
 
-| Prop                         | Type   | Required |
-| ---------------------------- | ------ | -------- |
-| [`children`](#menu-children) | `node` | false    |
+| Prop                                             | Type      | Required |
+| ------------------------------------------------ | --------- | -------- |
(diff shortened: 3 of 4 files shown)


chaance:
@thiniulian Do you mind reproducing the issue first in a code sandbox so I can have a look? I think this is probably something that can be handled in user code, which I'd much prefer over adding a prop. If not it may also just be something we do by default, but I am not seeing this issue when using our default positioning logic for the popover.

thiniulian:
Sure! I'll do it right away. 
The key is the window height being smaller than the height of the menu dropdown. My investigation turned to the `focus(menuRef.current);` being the culprit. Basically the browser scrolls to the element that becomes focused.

thiniulian:
Here you have a basic example. I created a huge list on purpose to make sure that there's a dropdown with a height larger than the window height.

https://codesandbox.io/s/stupefied-zhukovsky-3ttfh

packages/menu-button/src/index.tsx
```diff
@@ -145,7 +149,13 @@ export const Menu: React.FC<MenuProps> = ({ id, children }) => {
       // @ts-ignore
       window.__REACH_DISABLE_TOOLTIPS = true;
       window.requestAnimationFrame(() => {
-        focus(menuRef.current);
+        if (keepScrollPosition) {
```
indiesquidge, line 152: We need to include this in the dependency array.

indiesquidge:
@thiniulian, I think I've found a way to accomplish the scroll prevention in user code. What do you think about wrapping `MenuList` in a custom component like this?

```jsx
function MyCustomMenuList(props: MenuListProps) {
  const { isExpanded } = useMenuButtonContext();
  const [scrollOptions, setScrollOptions] = React.useState({
    left: window.scrollX,
    top: window.scrollY,
  });

  React.useEffect(() => {
    if (isExpanded) {
      // setTimeout forces scroll to occur *after* Reach UI focuses the menuRef
      // src: https://github.com/reach/reach-ui/blob/eb8773b897218d970d04b96923f61f88d9c5db07/packages/menu-button/src/index.tsx#L138-L140
      setTimeout(() => {
        window.scrollTo(scrollOptions);
      });
    }
  }, [scrollOptions, isExpanded]);

  React.useEffect(() => {
    const onScroll = () => {
      // track last scroll position before menu list expands so we know where to reset focus.
      if (!isExpanded) {
        const { scrollX, scrollY } = getScrollPosition();
        setScrollOptions({ left: scrollX, top: scrollY });
      }
    };

    window.addEventListener("scroll", onScroll);

    return () => {
      window.removeEventListener("scroll", onScroll);
    };
  }, [isExpanded]);

  return <MenuList {...props} />;
}
```

You could even pull out the event listener code into its own [custom hook][1] if you like.

Here's a [live example][2] of this working.

[1]: https://dev.to/n8tb1t/tracking-scroll-position-with-react-hooks-3bbj
[2]: https://codesandbox.io/s/gallant-bassi-p9tx0

thiniulian:
@indiesquidge thanks for the solution! 
This snippet of code works indeed, but I see a few issues:
- it will increase the complexity of my code
- there is a small flicker that happens, my hunch is that it's due to the setState and the render that this is triggering.

packages/menu-button/src/index.tsx
thiniulian: I added it. Thanks for pointing this out!

packages/menu-button/src/index.tsx
```diff
@@ -136,7 +141,13 @@ export const Menu: React.FC<MenuProps> = ({ id, children }) => {
...
+          const scrollOptions = { left: window.scrollX, top: window.scrollY };
+          focus(menuRef.current);
+          window.scrollTo(scrollOptions);
+        } else {
+          focus(menuRef.current);
+        }
```
indiesquidge, line 150: I think we could also clean this up slightly by only having the scrollTo behind a conditional

```js
const scrollOptions = { left: window.scrollX, top: window.scrollY };

focus(menuRef.current);

if (keepScrollPosition) {
  window.scrollTo(scrollOptions);
}
```

website/src/pages/menu-button.mdx
```diff
@@ -133,6 +134,13 @@ function Example() {
 }
 ```
 
+##### Menu `keepScrollPosition`
+
+`keepScrollPosition: boolean`
+
+When a menu is open the focus is trasferred to the menu itself. In the case the menu list element height is larger than the window height this focus will scroll the page making it jumpy.
+Passing `true` on this prop will still focus on the menu element but keeping the position of the scroll on the page.
```
indiesquidge, line 142: A few grammatical edits and spell checks:

```
When a `<Menu>` is opened the focus is transferred to the `<MenuList>`. If the `<MenuList>` height is longer than the viewport height, the focus will scroll the page so that the top of the `<MenuList>` is flush with the top of the viewport, which may result in a visual "jump".
Passing `true` for this prop will still focus the `<MenuList>`, but avoid the scroll position change, regardless of the `<MenuList>` height.
```

packages/menu-button/examples/avoid-page-scroll-on-focus.example.js
```diff
@@ -0,0 +1,36 @@
...
+
+let name = "Avoid Page Scroll on Menu Focus";
+
+function Example() {
+  const menus = [1, 2, 3, 4];
+  const links = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+  return (
+    <div>
+      <h2>Home</h2>
+      {menus.map((mainIndex) => (
```
indiesquidge, line 13: What is the reason for creating multiple menus for this example?

indiesquidge:
@thiniulian, glad it worked.

Regarding increasing the complexity of your code, I'm not sure what to say about that. Complexity has to live somewhere—if it's not in the app developer's code, it's in some library code. Unless the change is to address a popular recurring problem that cannot be solved elsewhere, I'm hesitant to have that complexity live in a library.

For this particular problem, the [example above][1] (or some variant of it) could work as a solution. Another solution to consider is having a fixed height on your menu list and allowing its contents of options to overflow scroll. This not only avoids the jumpy scroll behavior, but preserves the feature of the popover to appear on different sides of the menu button if there is not room below it, and also saves inundating a user with a list of options longer than their viewport. (Also if the number of options is really that large, consider using a [Combobox][2] instead to allow users to filter the list.)

I'll leave it to @chaance to decide if increasing the API footprint of this component is something we'd like to do. The small flicker with my example solution may well indeed mean that—if the scroll prevention is a behavior we'd like to support—the logic needs to happen within the component.

[1]: https://github.com/reach/reach-ui/pull/692#issuecomment-730607342
[2]: https://reach.tech/combobox

packages/menu-button/src/index.tsx
```diff
@@ -136,7 +141,13 @@ export const Menu: React.FC<MenuProps> = ({ id, children }) => {
       // @ts-ignore
       window.__REACH_DISABLE_TOOLTIPS = true;
       window.requestAnimationFrame(() => {
-        focus(menuRef.current);
+        if (keepScrollPosition) {
+          const scrollOptions = { left: window.scrollX, top: window.scrollY };
+          focus(menuRef.current);
```
chaance, line 146: Curious, would `menuRef.current.focus({ preventScroll: true })` not work here?

Issue #231, opened 2019-07-11
Tooltip stuck when surrounding a disabled button


MayaUribe:
When adding a `Tooltip` on top of a disabled `button`, the Tooltip gets stuck, I'm not sure if I'm missing a setting, here is an example of what I'm doing 😄

```
<Tooltip label="Notifications">
   <button style={buttonStyle} disabled={isDisabled}>
     Click Me
   </button>
</Tooltip>
```

![Screen Recording 2019-07-11 at 03 31 PM](https://user-images.githubusercontent.com/988972/61055575-1a4a4980-a3f2-11e9-8dcc-5d2a39fb610d.gif)

balazssagi:
I think it's related to disabled buttons not firing mouse events. As a workaround you can wrap the `button` with an other element.

venikx:
Can you send me the browser, version + OS?
At least for me I can't reproduce it on the latest Firefox (Nightly), but I can reproduce it for Google Chrome.

audiolion:
> I think it's related to disabled buttons not firing mouse events. As a workaround you can wrap the button with an other element.

sounds right, however wrapping the button in a `div` is not fixing the issue for me in Google Chrome

raunofreiberg:
Any workarounds for this? Wrapping the button does not work for me either. This seems like a pretty common use case for Tooltip (indicating disabled button reasoning)

raunofreiberg:
@chancestrickland Sorry to ping, but lately I noticed that more attention is being put towards the development of Reach UI, hence I'm wondering if this issue is high in the priority list? Thanks a lot for spending time on this, I really enjoy using Reach as a foundation :)

chancestrickland:
@raunofreiberg It's in the backlog but we're working through the issues. I'll update as soon as I've merged. If you could help out with a PR or even a CodeSandbox reproduction that would certainly get it moving faster!

raunofreiberg:
@chancestrickland Thanks for the response! I've set up a Codesandbox with a reproduction of the issue -> https://codesandbox.io/s/reach-ui-tooltip-disabled-button-issue-unwi0

venikx:
@raunofreiberg @chancestrickland  Confirmed to be broken on Chromium based browsers. On Firefox it doesn't get stuck. Chromium, Google Chrome and Vivaldi are broken.

chancestrickland:
[...]

Issue #564, opened 2020-04-27
[Tooltip] Use Pointer Events


raunofreiberg:
### Current Behavior

Using `@reach/tooltip` with disabled buttons is not possible. See: #231 

This is not specific to Reach UI, but rather a browser implementation detail, which can't be worked around without adding an additional wrapper around disabled buttons (which also has styling implications and is generally unpredictable).

### Desired Behavior

Using `@reach/tooltip` with disabled buttons would be _somewhat_ possible.

### Suggested Solution

I have experimented with using [Pointer Events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) (not to be mistaken with CSS `pointer-events` property) and they seem to produce desired results. 

More specifically, **`onPointerEnter` and `onPointerLeave` seem to both be fired on disabled buttons.** An example: https://codesandbox.io/s/gracious-galileo-qymqw?file=/src/App.js

I don't know what other implications this may have, or why do these events fire on disabled buttons, but it might be worth considering using said events instead of `onMouseEnter`/`onMouseLeave`, especially since `onMouseEnter` should not fire on disabled inputs as of [React 16.13.0](https://github.com/facebook/react/pull/17675)

### Who does this impact? Who is this for?

All users of Tooltip.

### Describe alternatives you've considered

I know this can be done in user land with the `useTooltip` hook by mapping the mouse events to pointer events, and I'm not expecting Reach UI to do anything here — but I thought it would be something interesting to consider. I think it would be really cool to support disabled buttons out of the box (which is a surprisingly common use case).

chancestrickland:
Interesting, thanks for bringing this up. Have you tested this on touch by chance? I'm a little concerned about `pointerenter`:

> The `pointerenter` event fires when a pointing device is moved into the hit test boundaries of an element or one of its descendants, **including as a result of a `pointerdown` event from a device that does not support hover**.

I'm not sure if we want tooltips showing up on touch events. I think some stylus inputs have a detectable hover range, but for those that don't we'd probably need to account for that. I'm open to a PR that uses pointer events only when supported and falls back to mouse events, but I'd also want it to be tested well on touch devices. I can't promise I'll have time to deal with them myself right away, but I'm happy to help test if you want to get it started!

raunofreiberg:
> Interesting, thanks for bringing this up. Have you tested this on touch by chance? I'm a little concerned about pointerenter:

[...]

PR #712, opened 2020-12-06
fix(tooltip): consistent cross-browser behavior with disabled elements

taifen:

This pull request:

- [ ] Creates a new package
- [ ] Other

----

Thanks for the great library!

This PR fixes https://github.com/reach/reach-ui/issues/231 and https://github.com/reach/reach-ui/issues/564

We use `@reach/tooltip` with disabled buttons and inputs and the need to create a wrapper around such components is rather inconvenient.

Pointer events API is [widely supported](https://caniuse.com/pointer) and PR contains a fallback just in case. 

Unfortunately couldn't update component tests because there is no support in `jsdom` yet https://github.com/jsdom/jsdom/pull/2666, I tried to use the proposed workarounds but they didn't work for me.  So tests actually check that component still works if there is no pointer events support.

It seems that Safari has incorrect implementation and triggers only `onPointerEnter` so I added a workaround for that. Checked on BrowserStack in IE11, Edge, Firefox, Chrome, and mobile Safari/Chrome and didn't notice a negative impact from that workaround.

### DIFF
--- a/packages/tooltip/src/index.tsx
+++ b/packages/tooltip/src/index.tsx
@@ -61,6 +61,7 @@ import PropTypes from "prop-types";
 
 const MOUSE_REST_TIMEOUT = 100;
 const LEAVE_TIMEOUT = 500;
+const POINTER_TYPE_MOUSE = "mouse";
 
 ////////////////////////////////////////////////////////////////////////////////
 // States
@@ -227,13 +228,17 @@ function clearContextId() {
  */
 function useTooltip<T extends HTMLElement>({
   id: idProp,
+  onPointerEnter,
+  onPointerMove,
+  onPointerLeave,
+  onPointerDown,
   onMouseEnter,
   onMouseMove,
   onMouseLeave,
+  onMouseDown,
   onFocus,
   onBlur,
   onKeyDown,
-  onMouseDown,
   ref: forwardedRef,
   DEBUG_STYLE,
 }: {
@@ -285,12 +290,73 @@ function useTooltip<T extends HTMLElement>({
     return () => ownerDocument.removeEventListener("keydown", listener);
   }, []);
 
+  React.useEffect(() => {
+    /*
+      This is a workaround for using tooltips with disabled controls in Safari.
+      Safari fires `pointerenter` but does not fire `pointerleave`
+      and `onPointerEventLeave` added to the trigger element will not work
+    */
+
+    if (!("PointerEvent" in window)) return;
+
+    let ownerDocument = getOwnerDocument(ownRef.current)!;
+
+    function listener(event: MouseEvent) {
+      // @ts-ignore `disabled` does not exist on HTMLDivElement but tooltips can be used with different elements
+      if (state !== VISIBLE || !ownRef.current?.disabled) return;
+
+      let target = event.target as Element | null;
+      if (
+        (target?.hasAttribute("data-reach-tooltip-trigger") &&
+          target?.hasAttribute("aria-describedby")) ||
+        target?.closest("[data-reach-tooltip-trigger][aria-describedby]")
+      ) {
+        return;
+      }
+
+      transition(GLOBAL_MOUSE_MOVE);
+    }
+
+    ownerDocument.addEventListener("mousemove", listener);
+    return () => ownerDocument.removeEventListener("mousemove", listener);
+  }, []);
+
+  function wrapMouseEvent<EventType extends React.SyntheticEvent | Event>(
+    theirHandler: ((event: EventType) => any) | undefined,
+    ourHandler: (event: EventType) => any
+  ) {
+    // Use internal MouseEvent handler only if PointerEvent not supported
+    if ("PointerEvent" in window) return theirHandler;
+
+    return wrapEvent(theirHandler, ourHandler);
+  }
+
+  function wrapPointerEventHandler(
+    handler: (event: React.PointerEvent) => any
+  ) {
+    return function onPointerEvent(event: React.PointerEvent) {
+      // Handle pointer events only from mouse device
+      if (event.pointerType !== POINTER_TYPE_MOUSE) return;
+      handler(event);
+    };
+  }
+
   function handleMouseEnter() {
     transition(MOUSE_ENTER, { id });
   }
 
   function handleMouseMove() {
     transition(MOUSE_MOVE, { id });
+  }
+
+  function handleMouseLeave() {
+    transition(MOUSE_LEAVE);
+  }
+
+  function handleMouseDown() {
+    // Allow quick click from one tool to another
+    if (context.id !== id) return;
+    transition(MOUSE_DOWN);
   }
 
   function handleFocus() {
@@ -301,20 +367,10 @@ function useTooltip<T extends HTMLElement>({
     transition(FOCUS, { id });
   }
 
-  function handleMouseLeave() {
-    transition(MOUSE_LEAVE);
-  }
-
   function handleBlur() {
     // Allow quick click from one tool to another
     if (context.id !== id) return;
     transition(BLUR);
-  }
-
-  function handleMouseDown() {
-    // Allow quick click from one tool to another
-    if (context.id !== id) return;
-    transition(MOUSE_DOWN);
   }
 
   function handleKeyDown(event: React.KeyboardEvent<T>) {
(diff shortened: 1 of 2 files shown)


taifen:
Hi @chaance. Could you take a look, please?

raunofreiberg:
Some context: https://github.com/reach/reach-ui/issues/564#issuecomment-620382316

As an alternative, you could drop down to the `useTooltip` hook and have your own Tooltip component use pointer events, which is what I do right now 🙃

Something like this would work:

```tsx
import { TooltipPopup, useTooltip } from '@reach/tooltip';

function Tooltip() {
  const [trigger, tooltip, isVisible] = useTooltip();

  const { onMouseEnter, onMouseLeave, ...triggerProps } = trigger;

  return (
    <>
      {React.cloneElement(
        children,
        children.props.disabled
          ? {
              onPointerEnter: onMouseEnter,
              onPointerLeave: onMouseLeave,
              ...triggerProps,
            }
          : trigger,
      )}
      {isVisible && <TooltipPopup />}
    </>
  );
}
```

taifen:
> Some context: [#564 (comment)](https://github.com/reach/reach-ui/issues/564#issuecomment-620382316)
> 
> As an alternative, you could drop down to the `useTooltip` hook and have your own Tooltip component use pointer events, which is what I do right now 🙃
> 
> Something like this would work:
> 
> ```tsx
> import { TooltipPopup, useTooltip } from '@reach/tooltip';
> 
> function Tooltip() {
>   const [trigger, tooltip, isVisible] = useTooltip();
> 
>   const { onMouseEnter, onMouseLeave, ...triggerProps } = trigger;
> 
>   return (
>     <>
>       {React.cloneElement(
>         children,
>         children.props.disabled
>           ? {
>               onPointerEnter: onMouseEnter,
>               onPointerLeave: onMouseLeave,
>               ...triggerProps,
>             }
>           : trigger,
>       )}
>       {isVisible && <TooltipPopup />}
>     </>
>   );
> }
> ```

Thanks! I need to support Safari though, so I use a forked version for now

packages/tooltip/src/index.tsx
```diff
@@ -285,14 +290,74 @@ function useTooltip<T extends HTMLElement>({
...
+    let ownerDocument = getOwnerDocument(ownRef.current)!;
+
+    function listener(event: MouseEvent) {
+      // @ts-ignore `disabled` does not exist on HTMLDivElement but tooltips can be used with different elements
+      if (state !== VISIBLE || !ownRef.current?.disabled) return;
+
+      let target = event.target as Element | null;
+      if (
+        target?.hasAttribute("data-reach-tooltip-trigger") ||
+        target?.closest("[data-reach-tooltip-trigger][aria-describedby]")
```
chaance, line 311: Can you explain the addition of `[aria-describedby]` in the selector here?

packages/tooltip/src/index.tsx
```diff
@@ -285,14 +290,74 @@ function useTooltip<T extends HTMLElement>({
...
+      and `onPointerEventLeave` added to the trigger element will not work
+    */
+
+    if (!("PointerEvent" in window)) return;
+
+    let ownerDocument = getOwnerDocument(ownRef.current)!;
+
+    function listener(event: MouseEvent) {
+      // @ts-ignore `disabled` does not exist on HTMLDivElement but tooltips can be used with different elements
+      if (state !== VISIBLE || !ownRef.current?.disabled) return;
```
chaance, line 306: Instead of checking the ref's element here, have you considered accepting `disabled` as an additional prop for the hook? That would allow us to use it as a dependency for the effect and avoid attaching useless event listeners for non-disabled components.

packages/tooltip/src/index.tsx
```diff
@@ -285,14 +290,74 @@ function useTooltip<T extends HTMLElement>({
...
+      handler(event);
+    };
+  }
+
+  function handleMouseOrPointerEnter() {
     transition(MOUSE_ENTER, { id });
   }
 
-  function handleMouseMove() {
+  function handleMouseOrPointerMove() {
```
chaance, line 347: Nit here, but can we keep all of these function names `handleMouse` instead of `handleMouseOrPointer`? I recognize why you changed these, but considering we're still only acting on mouse events I think the intent is better communicated as-is.

chaance (changes requested):
This looks good to me, just a few questions and requests. Thanks so much for the time here!

packages/tooltip/src/index.tsx
taifen: We can have multiple disabled elements (icon buttons, for example) with tooltips next to each other, so I wanted to make sure that `GLOBAL_MOUSE_MOVE` will be triggered if you quickly move to the next disabled element. And I missed `aria-describedby` in the first part of the condition initially. 

At first, I wanted to use `ownRef.current?.isEqualNode(target) || ownRef.current?.contains(target)` instead, but that didn't work and I'm not sure why.

packages/tooltip/src/index.tsx
taifen: ✅  Updated

packages/tooltip/src/index.tsx
taifen: Not sure if I understood correctly. Do you mean something like:
```tsx
<Tooltip id="wow" label="Notifications" isTriggerDisabled={isDisabled}>
  <Button isDisabled={isDisabled} />
</Tooltip>
```
?

Since this whole event listener is a workaround only for one browser I wouldn't add any additional prop to the hook or component. One more event listener in addition to another 4 doesn't make a huge impact and it's created only once. And it can be safely removed when/if the issue will be fixed in Safari.

chaance: Yeah it's a bit awkward as for the component API, but I think it makes a little more sense for the hook API. It would compose pretty naturally on buttons since `disabled` is a built-in prop for JSX (less-so if you have your own button API with something like `isDisabled`, but still). I'd be comfortable leaving it out of the component for the time being and considering that API separately while supporting the core logic in the hook.

```js
const CustomTooltipButton = React.forwardRef(({ label, ...props }, ref) => {
  // as long as `disabled` is a prop, this should work without changes to the existing API
  let [trigger, tooltip] = useTooltip({ ...props, ref });
  return (
    <React.Fragment>
      <button {...props} {...trigger} />
      <CustomTooltipPopup label={label} {...tooltip} />
    </React.Fragment>
  );
});
```

chaance: Actually, I just realized since we're already using `cloneElement` in `Tooltip` this is absolutely trivial to solve there as well. If `disabled` is passed to the child this should all just work with the addition of one line:

```js
disabled: child.props.disabled
```

packages/tooltip/src/index.tsx
chaance: Ah I think that makes sense. That could be a buggy assumption though seeing as how you can arbitrarily pass `aria-describedby` as well. I might add a data selector to the trigger params. We already use `data-state` elsewhere, so maybe:

```js
data-state={isVisible ? 'tooltip-visible' : 'tooltip-hidden'}
```

That might make the selector a bit more reliable! Also, you can just get away with `ownRef.current.contains` I think, that should return true even if the element itself matches the query.

taifen: `ownRef.current.contains` doesn't work in this case, unfortunately. I assume because of ref mutations. 

But I like `data-state`, it's also more intuitive than the previous selector. Thanks!  Also, `closest` returns the element itself so I removed checks for attributes on the target.

packages/tooltip/src/index.tsx
taifen: Pushed an update.

I'm a bit concerned about the meaning of `disabled` param here though. Knowing implementation details I understand that it belongs to trigger element params. Otherwise, I would expect the tooltip itself to be disabled when `disabled: true` is passed to `useTooltip`

chaance: Yeah that's a good point too. The obvious alternatives aren't clicking for me as good APIs. Let me think about it a bit, but if you have any other ideas I'm open. My main thing is, if we use the DOM as the source of truth we still need it to be reactive, so the effect would probably need a mutation observer or something.

chaance: The more I think about this, the less concerned I am about the distinction between a disabled trigger and a disabled tooltip. Here's why:

There really isn't a concept of a disabled tooltip IMO. [W3C describes the `aria-disabled` state](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-disabled) as an "element [that] is perceivable but disabled, so it is not editable or otherwise operable".

A tooltip is never really operable. It's not a control, but meant to provide supplementary content that is never supposed to be interactive anyway. Presumably if it were disabled, it wouldn't pop up at all which means its content would no longer be perceivable.

If for whatever reason someone decided that they had a valid use case for turning off a tooltip, they should probably compose their own with `useTooltip` and render it conditionally. Using `disabled` to describe a tooltip seems like it would always be semantically wrong.

Lastly, the other props accepted by `useTooltip` are composed by the trigger. Props intended for the tooltip itself can be passed directly to `TooltipPopup`. So to me it makes sense that `disabled` would be passed to the trigger in this case.

chaance (approved):

packages/tooltip/src/index.tsx
```diff
@@ -285,6 +292,55 @@ function useTooltip<T extends HTMLElement>({
...
+    ownerDocument.addEventListener("mousemove", listener);
+    return () => ownerDocument.removeEventListener("mousemove", listener);
+  }, [disabled]);
+
+  function wrapMouseEvent<EventType extends React.SyntheticEvent | Event>(
+    theirHandler: ((event: EventType) => any) | undefined,
+    ourHandler: (event: EventType) => any
+  ) {
+    // Use internal MouseEvent handler only if PointerEvent not supported
+    if ("PointerEvent" in window) return theirHandler;
```
karlhorky, line 329: Hmm... seems like this line is causing Gatsby to fail on server-side rendering:

```
$ cross-env NODE_ENV=production gatsby build
failed Building static HTML for pages - 0.750s
error "window" is not available during server side rendering.

  246 |   function wrapMouseEvent(theirHandler, ourHandler) {
  247 |     // Use internal MouseEvent handler only if PointerEvent is not supported
> 248 |     if ("PointerEvent" in window) return theirHandler;
      | ^
  249 |     return wrapEvent(theirHandler, ourHandler);
  250 |   }
  251 | 


  WebpackError: ReferenceError: window is not defined
  
  - tooltip.esm.js:248 
    /[@reach]/tooltip/dist/tooltip.esm.js:248:1
  
  - tooltip.esm.js:333 
    /[@reach]/tooltip/dist/tooltip.esm.js:333:1
  

not finished Generating image thumbnails - 147.857s
```

Reported in https://github.com/reach/reach-ui/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

PR #732, opened 2021-01-26
Replace custom build scripts w/ Preconstruct

chaance:
This PR replaces the current build scripts, which were forked and simplified based on tsdx, with [Preconstruct](https://preconstruct.tools). The tradeoffs of this change as I see it are:

Pros:
* Significantly faster build -- the entire repo builds in less than 20s now
* Similar build output -- we use almost exactly the same babel transforms as before, and preconstruct builds with rollup under the hood
* Less complexity to manage internally
* Some DX niceties to automate some tasks when creating new packages
* No more need for custom TS path aliases for each package (links for the dev env are created by `preconstruct dev`
* Preconstruct's source is familiar since it builds on top of rollup, so if we need to patch or fork for any reason it shouldn't be that scary

Cons:
* No more granular control over the build unless we want to fork the project
* Ummm...I think that's it?
* Really, pretty sure that's it

This pull request:

- [ ] Creates a new package
- [x] Other


chaance:
@Andarist this change is inspired by your mentioning of this project at one point, so if you have any thoughts or suggestions for effectively working within Preconstruct's constraints I'd be glad to hear them. I've already found myself itching for a few things here that I can no longer access, but I'm trying to convince myself it isn't worth it 😛

Andarist:
I like how you have summed up the cons :trollface: If you ever find a need to fork - let's discuss it first :)

babel.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  plugins: [
+    "babel-plugin-annotate-pure-calls",
```
Andarist, line 3: oh hello, hello my old friend 🤣

babel.config.js
```diff
@@ -0,0 +1,21 @@
...
+    "babel-plugin-dev-expression",
+    ["@babel/plugin-proposal-class-properties", { loose: true }],
+    "@babel/plugin-proposal-optional-chaining",
+    "@babel/plugin-proposal-nullish-coalescing-operator",
```
Andarist, line 7: those two are included by default now in `@babel/preset-env` as far as I remember - since they have been included in the language spec during Babel 7's lifetime

package.json
```diff
@@ -10,26 +10,30 @@
...
-    "build:pkg": "ts-node --transpile-only scripts/build-package",
+    "build": "yarn build:packages",
+    "build:packages": "preconstruct build",
     "ver": "lerna version --no-push --exact",
     "changes": "dotenv lerna-changelog",
     "clean": "git clean -e '!/.env' -e '!/website-deploy-key' -e '!/website-deploy-key.pub' -fdX .",
     "lint": "eslint .",
+    "fix": "preconstruct fix",
+    "dev": "yarn start",
+    "postinstall": "preconstruct dev",
```
Andarist, line 25: this pkg provides a nice & simple check so that your shared deps are always declared with the same dep range, highly recommend: https://github.com/Thinkmill/manypkg

Andarist: I mean - you could use it as part of postinstall script, like here: https://github.com/davidkpiano/xstate/blob/f36c5f84028e389bfb62e469f47253b7bce34d4f/package.json#L21

packages/accordion/package.json
```diff
@@ -9,25 +9,26 @@
...
+    "prop-types": "^15.7.2",
     "tslib": "^2.0.0"
   },
   "peerDependencies": {
     "react": "^16.8.0 || 17.x",
     "react-dom": "^16.8.0 || 17.x"
   },
-  "main": "dist/index.js",
-  "module": "dist/accordion.esm.js",
-  "typings": "dist/index.d.ts",
```
Andarist, line 27: maybe you would be willing to help with this issue: https://github.com/preconstruct/preconstruct/issues/259 ? :>

packages/alert/package.json
```diff
@@ -22,11 +19,14 @@
     "react": "^16.8.0 || 17.x",
     "react-dom": "^16.8.0 || 17.x"
   },
-  "main": "dist/index.js",
-  "module": "dist/alert.esm.js",
-  "typings": "dist/index.d.ts",
+  "main": "dist/reach-alert.cjs.js",
```
Andarist, line 22: just checking: were u using `exports: 'named'` for CJS output or are you sticking to the named exports only?

chaance: I believe I'm exposing everything as a named export and pretty sure I've updated the docs accordingly, but we had some default exports before that I've kept around for back compat. Is there something I need to be looking for that might break this?

chaance: Ah sorry you meant specifically for CJS -- yes, I previously used `exports: "named"` in rollup.

Andarist: Cool, I just have wanted to make sure that using Preconstruct didn't change the exported CJS shape for your packages - which could happen for `exports: 'auto'` (default) and lone default exports. 👍

Andarist (approved):
Excellent! Spread that Preconstruct love ❤️

PR #739, opened 2021-02-03
[skip-nav] Fix skip-nav tests

richiemccoll:
This PR updates the skip-nav tests to use [userEvent.tab](https://testing-library.com/docs/ecosystem-user-event/#tabshift-focustrap). This recreates the same behaviour that we see in the browser. 

---------

- [X] Add or edit tests to reflect the change (Run with `yarn test`).

This pull request:

- [ ] Creates a new package
- [X] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/skip-nav/__tests__/skip-nav.test.tsx
+++ b/packages/skip-nav/__tests__/skip-nav.test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { render /* act, fireEvent */ } from "$test/utils";
+import { render, userEvent, act, fireEvent } from "$test/utils";
 import { axe } from "jest-axe";
 import { SkipNavLink, SkipNavContent } from "@reach/skip-nav";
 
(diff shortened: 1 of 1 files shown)


packages/skip-nav/__tests__/skip-nav.test.tsx
```diff
@@ -11,23 +11,26 @@ describe("<SkipNavLink />", () => {
...
+      fireEvent.click(getByText("Skip Nav"));
+      userEvent.tab({ focusTrap: mainContent });
+      expect(getByText("Focus me")).toHaveFocus();
+    });
+  });
+  it("should work with a custom ID", () => {
+    let { getByText, getByTestId } = render(<Layout skipNavId="whatever" />);
+    const mainContent = getByTestId("main");
+    act(() => {
+      userEvent.tab();
```
indiesquidge, line 28: I don't think the first tab on either test is necessary; the tests pass without it since testing-library can find the skip nav link without it showing.

Was the intention to imitate the experience of tabbing to focus the `SkipNavLink` to make it visually visible? Maybe an interesting approach to that would be to tab to show the `SkipNavLink`, then use the enter key rather than a click event to trigger.

packages/skip-nav/__tests__/skip-nav.test.tsx
```diff
@@ -11,23 +11,26 @@ describe("<SkipNavLink />", () => {
...
+      fireEvent.click(getByText("Skip Nav"));
+      userEvent.tab({ focusTrap: mainContent });
```
indiesquidge, line 30: I'm not sure I understand the use of the focus trap here, can you help me understand why it's being used in this context? From the docs I see

> components like `focus-trap-react` will not work with `userEvent.tab()` or jsdom. For that reason, the `focusTrap` option is available to let you ensure your user is restricted within a focus-trap.

Reading this makes me feel like the use of the focus trap is to "mock" instances where a focus trap would already be in use (e.g. Reach's [Dialog][1]), but that isn't the case with `skip-nav`.

[1]: https://github.com/reach/reach-ui/blob/c5234721d372e2ab2a7745475007605e78d18895/packages/dialog/src/index.tsx#L185

chaance: Yeah I'm a little lost on this one as well. This seems like we're doing something we shouldn't need to do in order to get the test to pass.

PR #740, opened 2021-02-04
[menu-button] Prevent extraneous updates when clicking outside closed menus

fabiosantoscode:

This pull request:

- [ ] Creates a new package
- [x] Other -- Prevents re-renders

If creating a new package:

(snipped)

----

I'm using potentially hundreds of menus in my project, and I've noticed that whenever I click anywhere, a lot of React re-renders occurred. With a small number of menus this doesn't matter, though.

This happens because the useReducer callback always returns a fresh object, even if it's deep-equal to the previously returned one. I checked what was being done to prevent re-renders for this reason, and I found many conditionals that check for the current state, and don't call `dispatch` if the state is already correct. So I've done this as well to prevent `document.addEventListener('mousedown')`.

I've verified that this works using the react profiler:

In the "At the Corners" storybook story, run the profiler, click a blank spot, and stop the profiler. You can see multiple re-renders:

image

After my patch, performing the same action doesn't cause any re-renders.

### DIFF
--- a/packages/menu-button/src/index.tsx
+++ b/packages/menu-button/src/index.tsx
@@ -890,6 +890,10 @@ const MenuPopover = forwardRefWithAs<MenuPopoverProps, "div">(
     const ref = useForkedRef(popoverRef, forwardedRef);
 
     React.useEffect(() => {
+      if (!isExpanded) {
+        return;
+      }
+
       let ownerDocument = getOwnerDocument(popoverRef.current)!;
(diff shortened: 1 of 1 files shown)


fabiosantoscode:
I didn't write a test because I'm not sure what's the preferred way of doing so. Since this project uses jest (and I assume JSDOM) I could mock document.addEventListener and make sure it's not called. Let me know if this is a good way to test and I'll get to it.

chaance (approved):

Issue #666, opened 2020-09-07
[Portal] Add possibility to specify the container in which the portal is rendered


Drogglbecher:
### Current Behavior

I want to implement the `MenuButton` within a component with fullscreen option. The current implementation of the portal only allows `children` as props and renders the popover menu as child of `body`.

This behaviour is implemented in the followig [CodeSandbox Template](https://codesandbox.io/s/strange-tdd-cqk89) (You have to open the app in a new window to ensure fullscreen is working)

### Desired Behavior

It would be nice to specify the component content the portal is rendered in and allow to access this for all components using a portal.

### Suggested Solution

I think a good way could be adding a prop to the `Portal` like the `portalContainer` prop of [blueprints popover component](https://blueprintjs.com/docs/#core/components/popover)

john-oliveoil:
A feature like this, to allow us to specify the container or specify the node where the portal should be rendered is essential for us.

Why not just have a prop called `node` that allows us to specify a custom DOM element for the portal (if not specified then it would still append to the body as now). Basically the same as [react-portal](https://github.com/tajo/react-portal#portal).

This additional prop should extend to the other Reach packages (e.g. reach/dialog).

Without this feature it makes it impossible for us to correctly use this library in combination with the ShadowDOM. We do like Reach but without this we have been considering dropping it in favor of other packages. So please do consider adding this in a future (hopefully soon) version.

selrond:
Already mentioned in #571 

Another use case: say you have notification toasters / whatever UI element / you need to ensure is always on top of everything else.

In its current implementation, `@react/dialog` (which usese `@reach/portal` under the hood) just always [appends portal node to `body`](https://github.com/reach/reach-ui/blob/09b4d0a6f65803645e768a99f18907e96be7d4d6/packages/portal/src/index.tsx#L35) making it always the last child of `body`, which in turn makes it impossible to layer anything else on top of it in a predictable way.

So the underlying issue is that `@react/portal` makes an assumption that it should always be the last element in the DOM when rendered.

PR #756, opened 2021-03-02
Add optional node container target to Portal

Warsaken:
This PR aims to solve the issue #666.

It introduces a new optional prop for Portal: `nodeContainer` that allows the user to specify the containing node to render the Portal in.

If this PR is approved, then we can create a follow up PR to pass this prop to other relevant packages.

---

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/portal/src/index.tsx
+++ b/packages/portal/src/index.tsx
@@ -20,7 +20,11 @@ import { createPortal } from "react-dom";
  *
  * @see Docs https://reach.tech/portal#portal
  */
-const Portal: React.FC<PortalProps> = ({ children, type = "reach-portal" }) => {
+const Portal: React.FC<PortalProps> = ({
+  children,
+  type = "reach-portal",
+  nodeContainer,
+}) => {
   let mountNode = React.useRef<HTMLDivElement | null>(null);
   let portalNode = React.useRef<HTMLElement | null>(null);
   let forceUpdate = useForceUpdate();
@@ -32,11 +36,12 @@ const Portal: React.FC<PortalProps> = ({ children, type = "reach-portal" }) => {
     // In that case, it's important to append to the correct document element.
     const ownerDocument = mountNode.current!.ownerDocument;
     portalNode.current = ownerDocument?.createElement(type)!;
-    ownerDocument!.body.appendChild(portalNode.current);
+    const container = nodeContainer || portalNode.current!.ownerDocument.body;
+    container.appendChild(portalNode.current);
     forceUpdate();
     return () => {
-      if (portalNode.current && portalNode.current.ownerDocument) {
-        portalNode.current.ownerDocument.body.removeChild(portalNode.current);
+      if (portalNode.current && container) {
+        container.removeChild(portalNode.current);
       }
     };
   }, [type, forceUpdate]);
@@ -64,6 +69,12 @@ type PortalProps = {
    * @see Docs https://reach.tech/portal#portal-type
    */
   type?: string;
+  /**
+   * Target DOM element to render the portal.
+   *
+   * @see Docs https://reach.tech/portal#portal-nodeContainer
+   */
+  nodeContainer?: HTMLElement;
 };
 
 if (__DEV__) {
--- a/packages/portal/src/index.tsx
+++ b/packages/portal/src/index.tsx
@@ -23,7 +23,7 @@ import { createPortal } from "react-dom";
 const Portal: React.FC<PortalProps> = ({
   children,
   type = "reach-portal",
-  nodeContainer,
+  container,
 }) => {
   let mountNode = React.useRef<HTMLDivElement | null>(null);
   let portalNode = React.useRef<HTMLElement | null>(null);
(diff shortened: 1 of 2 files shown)


packages/portal/src/index.tsx
```diff
@@ -20,7 +20,11 @@ import { createPortal } from "react-dom";
  *
  * @see Docs https://reach.tech/portal#portal
  */
-const Portal: React.FC<PortalProps> = ({ children, type = "reach-portal" }) => {
+const Portal: React.FC<PortalProps> = ({
+  children,
+  type = "reach-portal",
+  nodeContainer,
```
chaance, line 26: Can we rename this prop to simply `container`?

packages/portal/src/index.tsx
```diff
@@ -32,11 +36,12 @@ const Portal: React.FC<PortalProps> = ({ children, type = "reach-portal" }) => {
     // In that case, it's important to append to the correct document element.
     const ownerDocument = mountNode.current!.ownerDocument;
     portalNode.current = ownerDocument?.createElement(type)!;
-    ownerDocument!.body.appendChild(portalNode.current);
+    const container = nodeContainer || portalNode.current!.ownerDocument.body;
```
chaance, line 39: I think we can simplify this a tiny bit

```
let body = container || ownerDocument.body
body.appendChild(portalNode.current)
```

packages/portal/src/index.tsx
```diff
@@ -64,6 +69,12 @@ type PortalProps = {
    * @see Docs https://reach.tech/portal#portal-type
    */
   type?: string;
+  /**
+   * Target DOM element to render the portal.
+   *
+   * @see Docs https://reach.tech/portal#portal-nodeContainer
+   */
+  nodeContainer?: HTMLElement;
```
chaance, line 77: Let's make `container` of type `Node`, `Document` is the default and really is can be any Node type.

website/src/pages/portal.mdx
```diff
@@ -62,9 +62,11 @@ Renders content inside of a portal at the end of the DOM tree.
...
 #### Portal Props
 
-| Prop                    | Type | Required |
-| ----------------------- | ---- | -------- |
-| [`children`](#children) | node | true     |
+| Prop                              | Type        | Required |
+| --------------------------------- | ----------- | -------- |
+| [`children`](#children)           | node        | true     |
+| [`type`](#type)                   | string      | false    |
+| [`nodeContainer`](#nodeContainer) | HTMLElement | false    |
```
chaance, line 69: Update the prop name and type here as well!

website/src/pages/portal.mdx
```diff
@@ -83,3 +85,9 @@ Any content you want to render inside of the portal.
 _Type_: `string` default: `reach-portal`
 
 The DOM element type to render.
+
+##### Portal nodeContainer
+
+_nodeContainer_: `HTMLElement` default: `document.body`
+
+The DOM element container of the portal. If not set the portal will be appended to the body of the document.
```
chaance, line 93: Update the prop name and type here as well, and tweak to read "The DOM **node** container..."

chaance (changes requested):
Thanks for the PR! A few changes requested but I'd like to merge this when we can.

Warsaken:
@chaance thanks for reviewing, I made the requested changes now.

chaance:
On second thought I think it makes a lot more sense for this prop to accept a ref rather than a DOM node. Probably requested changes before I thought about it more fully, so sorry about that. Let's hold for right now, I'll circle back to this after I've taken care of a few high-priority bugs for the next release.

Warsaken:
>  On second thought I think it makes a lot more sense for this prop to accept a ref rather than a DOM node. Probably requested changes before I thought about it more fully, so sorry about that. Let's hold for right now, I'll circle back to this after I've taken care of a few high-priority bugs for the next release.

Several of the other libraries I encountered write it as an `HTMLElement` (e.g.: https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/portal/portal.tsx) that's why I started with that type, since it's familiar and easy to migrate existing code to Reach later. But let me know which type you think is the best here and I can make those changes.

Warsaken:
@chaance This has been here for a while so I went ahead and updated the type to a `ref` like you suggested.

Being of `node` type or `HTMLElement` also seemed reasonable to me, for example in my use case the container I need to append the portal to is obtained through a function that returns a `HTMLElement` which may not even be a react component.
But either way works, I just need to convert that to the ref format afterwards.

Let me know if there are any other changes necessary. Thanks.

zefj:
hey @chaance , we could really use this feature too. Our case involves `menu-button` within overflow: hidden nodes living within modals, so unfortunately we cannot just disable the portal :disappointed: Is there any chance of getting this soon?

chaance (approved):

Warsaken:
I see the PR was merged, thanks @chaance.

Now we need to create a PR to pass the props along for the other libraries that use Portal (e.g. Dialog).
Will you, or someone else, create a PR for that or should I do it?

PR #757, opened 2021-03-03
If Combobox openOnFocus is true then also open when the input is clicked.

ProggerPete:
I was surprised to discover that clicking on the input doesn't automatically open the ComboboxList.  This means that there is no way to make a selection using the mouse alone.

openOnFocus is enough for the 1st click, but after selecting a value the input retains focus, so the only way to open it again without using the keyboard is to click away and click back.

Reopening the list if the user clicks again on the input seems a sensible strategy.


chaance:
I'm not sure this is needed. Here is an example where this behaves as expected. Do you have a reproduced issue where this doesn't already work?

ProggerPete:
> I'm not sure this is needed. Here is an example where this behaves as expected. Do you have a reproduced issue where this doesn't already work?
> 
> https://codesandbox.io/s/infallible-chatelet-1qd37?file=/src/App.js

Your example shows exactly the behaviour I think is a problem.  The first time you click the input the list shows.  If you click one of the list options it is selected and the lists closes.  Intuitively, if I want to open the list again I will click again on the input.  Doing so does not open the list.

This has been discussed further here : https://github.com/reach/reach-ui/issues/758

The users I build for are far happier if they click to reopen the list and it certainly seems better to me, but perhaps we are unique.  Using a fork isn't too problematic so I'm just doing that.  However I think the component would be better if it supported this user story.

chaance:
Ah, I see now. Thanks for linking to the issue for more context, I must have missed or forgotten that one. I'll take another look at this soon!

chaance (approved):

PR #761, opened 2021-03-04
Add sideEffects false to all packages to enable tree shaking

hipstersmoothie:

This pull request:

- [ ] Creates a new package
- [ ] Other

### DIFF
--- a/packages/accordion/package.json
+++ b/packages/accordion/package.json
@@ -4,6 +4,7 @@
   "description": "Accessible React accordion component",
   "author": "React Training <hello@reacttraining.com>",
   "license": "MIT",
+  "sideEffects": false,
   "repository": {
     "type": "git",
     "url": "git+https://github.com/reach/reach-ui.git",
--- a/packages/alert-dialog/package.json
+++ b/packages/alert-dialog/package.json
@@ -4,6 +4,7 @@
   "description": "Accessible React Alert Dialog.",
   "author": "React Training <hello@reacttraining.com>",
   "license": "MIT",
+  "sideEffects": false,
   "repository": {
     "type": "git",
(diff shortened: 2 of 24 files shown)


charlie632:
How's utils being bundled at the end? I've noticed that Webpack has troubles tree shaking modules that are in a single file. For a project I'm working on I needed to add `sideEffects: false` flag and split a module into several files.

hipstersmoothie:
@charlie632 You are correct. I think a change in the build tool reach uses will be needed

https://github.com/preconstruct/preconstruct/issues/380

https://twitter.com/HipsterSmoothie/status/1367574853843255296

chaance:
I'm also open to splitting utils up into separate files/entrypoints and getting rid of the index file altogether. I generally like to keep files at a minimum, but in the case of shared utils that might be the simpler option.

hipstersmoothie:
I think we should actually be good to merge without splitting up the utils.

Andarist (approved):
Clarification: This will bring *small* benefits of dropping those modules if they *stay* completely unused by the consuming application. If any import will be used from any given package then the whole package will be included *and* dead-code elimination will be attempted and its content - so depending on that content things might be dropped or not.

I've checked the `@reach/utils` and just a few bytes can be added (for this package) unnecessarily: https://twitter.com/AndaristRake/status/1367756327699746819

chaance:
Appreciate all the research and work to improve this @hipstersmoothie @Andarist! I'm still probably going to split the `utils` package up into separate entrypoints and import them individually just to make tree-shaking more reliable for everyone, but this is a great starting point.

ykzts:
With this change, we have to add `sideEffects: true` to  CSS settings in webpack.config.js. I think it would be good to add the documentation to https://reach.tech/styling/.

PR #769, opened 2021-03-29
Add *.css files to the list of sideEffect(ful) files

Andarist:
This should resolve https://github.com/reach/reach-ui/pull/761#issuecomment-809307675

### DIFF
--- a/packages/accordion/package.json
+++ b/packages/accordion/package.json
@@ -4,7 +4,9 @@
   "description": "Accessible React accordion component",
   "author": "React Training <hello@reacttraining.com>",
   "license": "MIT",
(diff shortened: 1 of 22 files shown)


chaance:
I'm sure this is fine but I need to research a bit more for my own sake. I'll take care of this ASAP.

Andarist:
`sideEffects` flag is actually an allowlist where if you use `false` as the value you are saying that *no files* have side-effects. You can use patterns to include files though and this pattern can be seen in the official webpack docs: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

Issue #755, opened 2021-03-02
Combobox: Binding `value` undesirely expands combobox


bentolor:
## 🐛 Bug report

Thanks @chaance for this beautifuly library. We are trying to use it in a Form editor page and have an issue with the combobox

### Current Behavior

When we set the current value, the combobox displays the expand state including popup. This even occurs if it is not visible (i.e. being in a collapsed section). Regardless of that, it makes it unusable, as it overlays/hides many other fields in the form.

This is how we use it in React:

```
import { faChevronDown } from '@fortawesome/free-solid-svg-icons'
import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
import {
  Combobox,
  ComboboxInput,
  ComboboxList,
  ComboboxOption,
  ComboboxPopover,
} from '@reach/combobox'
import '@reach/combobox/styles.css'
import React from 'react'
import Attribute from './Attribute'
import Delete from './shared/Delete'

/**
 * @param {{
 *  label: string
 *  description: string
 *  options: string[]
 *  defaultValue?(): string
 *  required?: boolean
 *  deletable?: boolean
 *  validationErrors: import('../../../../shared/validationTypes').ValidationError[]
 *  dataPath: string
 *  value: unknown
 *  onUpdate({}): void
 * }} props
 */
export default function EnumAttribute({
  options,
  required = true,
  deletable,
  ...props
}) {
  return (
    <Attribute {...props}>
      {({ onChange, onDelete }) => (
        <div className="max-w-md flex">
          <div className="w-full">
            <Combobox
              className="w-full"
              openOnFocus
              onSelect={(item) => {
                onChange(item)
              }}
            >
              <label className="block w-full flex">
                <ComboboxInput
                  value={/** @type {string} */ (props.value)}
                  className="border border-gray-400 py-1 px-2 w-full shadow-inner rounded-l"
                  selectOnClick
                  required={required}
                  onChange={(e) => {
                    onChange(e.target.value)
                  }}
                />
                <div className="flex items-center justify-center w-8 text-xs border border-gray-400 rounded-r bg-white hover:bg-gray-200 cursor-pointer">
                  <FontAwesomeIcon icon={faChevronDown} />
                </div>
              </label>
              <ComboboxPopover>
                <ComboboxList persistSelection>
                  {options.map((option, index) => (
                    <ComboboxOption key={index} value={option} />
                  ))}
                </ComboboxList>
              </ComboboxPopover>
            </Combobox>
          </div>
[...]

PR #783, opened 2021-04-14
[combobox] Fix binding value undesirely expands combobox

dlacaille:
Closes https://github.com/reach/reach-ui/issues/755

This pull request fixes an issue where binding a controlled value undesirely expands combobox.

The change adds`inputValueChangedRef` to ComboboxInput and a new event type `SIMULATED_CHANGE`.

When the input triggers onChange, we set `inputValueChangedRef` to `true` and reset it to `false` after we detect a controlled value change. We then make call `CHANGE` only if this value is `true` when a change is detected, otherwise we call `SIMULATED_CHANGE`. This deprecates the `INITIAL_CHANGE` event type as it becomes a `SIMULATED_CHANGE`.

---

Thank you for contributing to Reach UI! Please 
fill in this template before submitting your PR to help us process your request more quickly.

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/combobox/examples/simulated-change.example.js
+++ b/packages/combobox/examples/simulated-change.example.js
@@ -1,0 +1,91 @@
+import * as React from "react";
+import {
+  Combobox,
+  ComboboxInput,
+  ComboboxList,
+  ComboboxOption,
+  ComboboxPopover,
+} from "@reach/combobox";
+import { useCityMatch } from "./utils";
+import "@reach/combobox/styles.css";
+
+let name = "Controlled";
+
+function Example() {
+  let [term, setTerm] = React.useState("Detroit");
+  let [selection, setSelection] = React.useState("");
+  let results = useCityMatch(term);
+  let ref = React.useRef();
+
+  const handleChange = (event) => {
+    setTerm(event.target.value);
+  };
+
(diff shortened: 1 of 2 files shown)


Rada8732 (approved):

adalinesimonian:
Really looking forward to this fix. I have a component that allows selecting named colours as well as using a colour picker. Every time the colour picker updates the combobox input the entire list of colours pops up - not exactly the behaviour one would expect.

pablo-abc:
Would it be correct to assume that the `Combobox`'s popover should _ONLY_ open when its `ComboboxInput` has focus? I had made a patch for this as well and later I saw this PR already existed. Our solutions where quite similar except on the fact that I used an `isFocusedRef`: https://github.com/pablo-abc/reach-ui/pull/1/files (I made a PR on my own fork to not clutter this repo).

Basically I'm only calling `transition(CHANGE, { value })` only if `isFocusedRef.current === true`. Is there a situation on which the popover might be needed to open when the input does _not_ have focus?

chaance:
This looks great, thanks @dlacaille! Added a test and updated the storybook index since we updated some of that stuff internally since this PR was created.

PR #812, opened 2021-06-25
[menu-button] Prevent default for links to avoid double click events

### LINKED ISSUE #811: [MenuLink] Immediately clicks button inside Dialog when using keyboard
### Current Behavior

When opening a `@reach/dialog` from a `<MenuLink as={Link}>` (react-router Link) using the keyboard, any button in the modal is immediately clicked.

In Chrome, Safari, and Firefox, using the enter key causes the button to be clicked.

In Firefox, using the spacebar causes the button to be clicked.

Adding `onClick={(e) => e.stopPropagation()}` does not solve the issue.

### Expected behavior

I would expect that activating a MenuLink with the keyboard would behave the same as clicking with the mouse, and should not trigger a click event on a button inside a modal.

### Reproducible example

https://codesandbox.io/s/friendly-beaver-1n8rn?file=/src/index.js

Note that opening the modal using the mouse works fine, but using the keyboard does not show the modal.  Check the console.log output to see the modal is immediately closing because the close button inside is being "clicked".

### Suggested solution(s)

It seems this is being caused by calling `.click()` [here](https://github.com/reach/reach-ui/blob/1f2d2eece122ff3315cd425c1945034dc016cd88/packages/menu-button/src/index.tsx#L666), without first calling `preventDefault` on the original event.

### Additional context

I'm working on a PR to address this.

### Your environment

| Reach Package    | MenuLink, Dialog        |    @reach/menu-button@0.15.1, @reach/dialog@0.15.0     |
| React            |         |    17.0.2     |
| Browser          | Chrome, Firefox, Safari        |    Latest?     |
| Assistive tech   |  Keyboard       |         |
| Node             |         |  14.16.1       |
| npm/yarn         |    npm     |  7.17.0       |
| Operating System |   macOS      |   11.4      |

IanVS:

- [X] Ensure formatting is consistent with the project's Prettier configuration.

This pull request:

- [ ] Creates a new package
- [X] Fixes a bug in an existing package
- [ ] Other

Fixes #811.  

I found that without preventing the original event's default, I was getting two click events, and one would fire inside a Dialog that I was opening, immediately closing it.  This ensures the original event is prevented when creating a new .click() on a link.

### DIFF
--- a/packages/menu-button/src/index.tsx
+++ b/packages/menu-button/src/index.tsx
@@ -662,10 +662,10 @@ const MenuItems = React.forwardRef(function MenuItems(
           // consistent behavior across menu items we'll trigger a click when
           // the spacebar is pressed.
           if (selected) {
+            event.preventDefault();
             if (selected.isLink && selected.element) {
               selected.element.click();
             } else {
-              event.preventDefault();
               // Focus the button first by default when an item is selected.
               // We fire the onSelect callback next so the app can manage
               // focus if needed.


IanVS:
Oh wow, the codesandbox bot is awesome.  It forked my reproduction from the issue and applied these changes?  That's very cool, and a good way to show that the problem is fixed.

packages/menu-button/src/index.tsx
```diff
@@ -662,10 +662,10 @@ const MenuItems = React.forwardRef(function MenuItems(
           // consistent behavior across menu items we'll trigger a click when
           // the spacebar is pressed.
           if (selected) {
+            event.preventDefault();
```
chaance, line 665: @IanVS Should this not be moved into the next `if` block? I don't think we want to `preventDefault` for buttons.

chaance: Ah nope, I'm totally wrong here since we were doing that already! Gonna poke around a bit for any unintended edge cases here.

PR #817, opened 2021-07-08
[MenuList]: Add position prop for MenuList

NicolaiF:

This pull request:

This simplifies usage with the MenuList component with regards to positioning the list, instead of using MenuPopover.


NicolaiF:
Should be pretty straight forward, as MenuList is already composed by MenuPopover which supports the position prop.

Please let me know if I missed anything!

chaance:
I'm not sure we really want to add this. The point of providing separate components is so that if you need to access the popover apart from the list, you can drop down to the lower level API and access each of them directly. Routing props from `MenuList` to either `MenuPopover` or `MenuItems` isn't super useful when consumers can just create their own `MenuList` and design whatever API they want with it!

chaance:
If you want to customize `MenuList`, my suggestion is to import `MenuPopover` and `MenuItems` and then wrap them as you'd like. In fact, I'd say there's a decent chance we may deprecate `MenuList` in a future version and make this the recommendation for everyone.

```tsx
import { MenuPopover, MenuItems } from "@reach/menu-button";
import type { Position } from "@reach/popover";
import type * as Polymorphic from "@reach/utils/polymorphic";

const MenuList = React.forwardRef(
  ({ portal = true, position, ...props }, forwardedRef) => {
    return (
      <MenuPopover position={position} portal={portal}>
        <MenuItems {...props} ref={forwardedRef} data-reach-menu-list="" />
      </MenuPopover>
    );
  }
) as Polymorphic.ForwardRefComponent<"div", MenuListProps>;

interface MenuListProps {
  portal?: boolean;
  position?: Position;
  children: React.ReactNode;
}
```

PR #827, opened 2021-08-09
Descendants: Fix a bug where new items get wrong index

btraut:

This pull request:

- [ ] Creates a new package
- [X] Fixes a bug in an existing package
- [ ] Other

---

I noticed when using the descendants package that `registerDescendant` is always getting called with a set index, never null. It looks to me as if the `set` method inside `registerDescendant` is meant to get `index === null` when a new element has been added and needs to be compared to its siblings in the DOM via `compareDocumentPosition`. Instead, the newest items all come in with `index: -1`, then get sorted to the front of the items array and treated as if they're the first descendants, even when they're not.

I admit that I wasn't able to properly build a test for this fix because I couldn't figure out how to set up the repo on my own machine. Fortunately, I was able to copy/paste the entire descendants file into my project and verify that the fix is working. Perhaps @chaance can help create a better test?

### DIFF
--- a/packages/descendants/src/index.tsx
+++ b/packages/descendants/src/index.tsx
@@ -64,7 +64,7 @@ function useDescendant<DescendantType extends Descendant>(
 
     registerDescendant({
       ...descendant,
-      index,
+      index: index === -1 ? null : index,
     } as DescendantType);
     return () => {
       unregisterDescendant(descendant.element);


btraut:
Looking at this a little closer, I'm more confident in this fix. As it stands today, `registerDescendant` will NEVER be called with `index: null` because the `indexProp` uses nullish coalescing just above the `registerDescendant(…)` call. If `index` is never null, it means that most of the `set` function starting on line 116 is dead code.

chaance:
> I admit that I wasn't able to properly build a test for this fix because I couldn't figure out how to set up the repo on my own machine. Fortunately, I was able to copy/paste the entire descendants file into my project and verify that the fix is working. Perhaps @chaance can help create a better test?

I'd be happy to help. Do you mind creating a reproduction of the problem in a Code Sandbox so I can see exactly what we're trying to fix in action?

myrjola:
Here's a reproduction of one bug that this PR fixes https://codesandbox.io/s/reach-ui-template-forked-il3x4?file=/src/App.tsx

1. Click two times on the "Add one more guest" button. This adds two more options to the listbox.
2. Click the listbox open, the "1 room" options should be selected
3. Press the "down arrow" key, I expected "2 rooms" to be selected, but instead "3 rooms" is selected.

Before pressing down arrow:
image

After pressing down arrow:
image

Expected result:
image

myrjola:
I tried adding test coverage for this bug. However, I failed to get the down arrow press to work in the test. This is how far I've come:

```
    it(`selects the next option in the listbox when pressing down`, () => {
      let { getByRole, getAllByText } = render(
        <DynamicContent />
      );

      // Increment to three guests, this adds two more selections to the listbox.
      act(() => {
        const incrementGuestsButton = getByRole("button", {name: "Add one more guest"})
        incrementGuestsButton.click();
        incrementGuestsButton.click();
      });

      act(() => void fireMouseClick(getByRole("button", { name: "Select rooms 1 room" })));

      function getOptionByText(text: string) {
        return getAllByText(text).find(
          (element) => element.getAttribute("role") === "option"
        );
      }
      expect(getOptionByText("1 room")).toHaveAttribute("aria-selected", "true");

      act(() => {
        fireEvent.keyDown(document.activeElement!, { key: "ArrowDown" });
      })
      expect(getOptionByText("2 rooms")).toHaveAttribute("aria-selected", "true");
    });
```
The test requires the following component defined:

```
function DynamicContent() {
  let [guestCount, setGuestCount] = React.useState(1);

  const options = Array.from({ length: guestCount }).map((_, i) => {
    const roomCount = i + 1;
    return {
      value: roomCount.toString(),
      label: `${roomCount} ${roomCount > 1 ? "rooms" : "room"}`,
    };
  });

  return (
    <div>
      <span>
        {guestCount} {guestCount > 1 ? "guests " : "guest "}
      </span>
      <button
        type="button"
        onClick={() => setGuestCount((guestCount) => guestCount + 1)}
      >
        Add one more guest
      </button>
      <hr />
      <VisuallyHidden id="room-label">Select rooms</VisuallyHidden>
      <Listbox aria-labelledby="room-label" defaultValue="1" portal={false}>
        {options.map((option) => (
          <ListboxOption
            key={option.value}
            value={option.value}
            label={option.label}
          >
            {option.label}
          </ListboxOption>
        ))}
      </Listbox>
    </div>
  );
}
```

Issue #543, opened 2020-04-09
Ability to pass targetRef to ComboboxPopover just like a normal Popover


sanfilippopablo:
### Current Behavior

The `ComboboxPopover` passes a hardcoded ref (the input) to the `targetRef` prop of `Popover`.

### Desired Behavior

I have an use case where I want the Combobox autocomplete list to attach to a different element than the input (just a wrapper around it) and grab that element's position/width instead. It'd be great to just pass a ref of that element to the popover.

### Suggested Solution

In `ComboboxPopover`, if `targetRef` is coming as a prop, pass that to `Popover` instead of the inputRef.

### Who does this impact? Who is this for?

I guess this is an advanced case?

### Describe alternatives you've considered

None.

### Additional context

N/A

indiesquidge:
@sanfilippopablo Recognizing that it could be valuable to allow for the targetRef to be passed to the ComboboxPopover as a prop, there is also an option to [render the listbox inline][1] by passing `portal={false}`, which would allow you to style the listbox to match your input wrapper's width.

[1]: https://reacttraining.com/reach-ui/combobox#comboboxpopover

indiesquidge:
@chancestrickland, is this something we'd like to support? I'd be happy to implement it. I've seen a few other cases come up with folks wanting to attach the popover to a wrapper element around the input.

<img width="50%" src="https://user-images.githubusercontent.com/3409645/80767051-d436b180-8afb-11ea-93ee-82e4b9699a56.png" />

chancestrickland:
Sorry for the delayed response here. You can position the popover however you'd like with the `position` prop in popover, but it does look like we don't let you override that for combobox. Would that not be sufficient to handle custom positioning logic? You could use any other DOM refs you'd like in that case, no need to accept a custom `target` ref.

sanfilippopablo:
@chancestrickland although not ideal, I guess it could work if I get a ref to my outer element in my component and then pass a `position` prop defined in the component that ignores the `targetRef` and uses that outer ref?

PR #845, opened 2021-10-07
Combobox: Respect targetRef at ComboboxPopover component

vovakulikov:
## Context 

Sometimes (pretty often) we have to change the target ref of the Combobox Popover element in case if we want to build something new with `@reach/combobox`. A good example of that can be a multi-Combobox with pills. I think this problem has been described already in this issue https://github.com/reach/reach-ui/issues/543. 

This PR fixes the problem when you can't pass a custom target ref to the `<ComboboxPopover />` explicitly.

<table>
 <tr>
   <th>With input as a ref</th>
   <th>With combobox wrap as a ref</th>
 </tr>
 
 <tr>
   <td>
   
   
Screenshot 2021-10-08 at 00 03 27

   </td>
   
   <td>
      
      
image

   </td>
 </tr>
</table>

This PR also updates tokens Combobox a little bit and shows new case usage with `targetRef` at `<ComboboxPopover />`

### DIFF
--- a/packages/combobox/examples/token-input.example.js
+++ b/packages/combobox/examples/token-input.example.js
@@ -18,6 +18,8 @@ function Example() {
   let [term, setTerm] = React.useState("");
   let [selections, setSelections] = React.useState([]);
   let results = useCityMatch(term);
+  let containerRef = React.useRef();
+  let [customTargetMode, setCustomTargetMode] = React.useState(false);
 
   const handleChange = (event) => {
     setTerm(event.target.value);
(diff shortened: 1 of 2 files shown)


indiesquidge:
@vovakulikov is there a reason for having a custom target ref other than for positioning the popover? If not, I've found that not using a portal, absolutely positioning the popover, and relatively positioning the element you want to align it with will achieve the behavior you're looking for as well.

If you still want to use the popover as a portal, as noted in #543, `ComboboxPopover` [now accepts][1] the `position` prop is of type [Position][2] and is passed on the underlying Popover. If you wanted to use your own target ref for the wrapping div for the `position` prop, you could write something like

```jsx
import { positionMatchWidth } from "@reach/popover";
import { useRect } from "@reach/rect";

function Example() {
  const wrapperRef = React.useRef<HTMLDivElement | null>(null);
  const wrapperRect = useRect(wrapperRef);

  return (
    <div style={{ padding: 20, border: "1px solid tomato" }} ref={wrapperRef}>
      <Combobox id="holy-smokes" aria-label="choose a city">
        <ComboboxPopover position={(_, popoverRect) => positionMatchWidth(wrapperRect, popoverRect)} />
      </Combobox>
    </div>
  );
}
```
Which comes out looking something like

image

[1]: https://github.com/reach/reach-ui/commit/7d6d7994a36aed1aae77450a910a6ecae2fb8ee5#diff-dd4c59e5bf44924a1cb62a519fc8af874f8e245e3b7ed635a61cd7f51dd8f7c2
[2]: https://github.com/reach/reach-ui/blob/d72226c86d2b4d201027556d51adc3ba9d2f8d87/packages/popover/src/index.tsx#L360-L364

vovakulikov:
@indiesquidge thanks. I've missed that approach with position with `useRect` instead of passing target ref. You're right it's possible and probably more flexible rather than  `targetRef`. I think it worth it to add this case to the Combobox description doc or at least to the storybook story. what do you think?

PR #863, opened 2021-11-07
Combobox: Up arrow in IDLE state now passes persistSelection to navigate, the same that down arrow does

kevinstubbs:
Tiny fix to make the up arrow key behave the same that the down arrow key does when in IDLE state.

- [YES ] Use a meaningful title for the pull request. Include the name of the package modified.
- [YES] Test the change in your own code (Compile and run).
- [NA] Add or edit tests to reflect the change (Run with `yarn test`).
- [NA] Add or edit Storybook examples to reflect the change (Run with `yarn start`).
- [YES] Ensure formatting is consistent with the project's Prettier configuration.
- [NA] Add documentation to support any new features.

This pull request: **Fixes a bug in an existing package**

**Not a new package.**

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -1079,7 +1079,10 @@ function useKeyDown() {
         }
 
         if (state === IDLE) {
-          transition(NAVIGATE);
+          // Opening a closed list
+          transition(NAVIGATE, {
+            persistSelection: persistSelectionRef.current,
+          });
         } else {
           let prev = getPreviousOption();
           transition(NAVIGATE, { value: prev ? prev.value : null });


dmamie:
+1 to merge this, blocking us as well

indiesquidge (approved):
Lgtm. For clarity for other reviewers, this affects the up arrow when the `persistSelection` prop is passed to the `ComboboxList`, making it so that the up/down arrow keys will both open the popover and highlight the matching selection.

| Before | After |
| :---: | :---: |
| before | after |

Notice in the "before" gif, down arrow will highlight the selected match upon opening the popover, but up arrow will not. The arrows behave the same in the "after" gif.

indiesquidge:
@dmamie we understand why this change might be useful and necessary, but this library does not currently have full-time support, so merges and new version deploys may be slow.

Keep in mind that you can always patch this behavior in the meantime in your own project with https://www.npmjs.com/package/patch-package

chaance (approved):

PR #865, opened 2021-11-12
Fix selection for controlled combobox

arackaf:
Here's the context. In the existing Storybook, go to combobox controlled (third one). Then repeatedly type in the box, and select something. Occasionally (roughly 1 in 5 times) you'll see the text of the thing you select flash in the input, before being cleared. To make this more pronounced, and easier to see, edit this code:

```js
    case SELECT_WITH_CLICK:
      return {
        ...nextState,
        value: event.value,
        navigationValue: null,
      };
```

to this

```js
    case SELECT_WITH_CLICK:
      return {
        ...nextState,
        value: "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
        navigationValue: null,
      };
```

This PR fixes that, by changing the value to null when the combobox is used in controlled mode. 

--- 

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/packages/combobox/src/index.tsx
+++ b/packages/combobox/src/index.tsx
@@ -313,7 +313,6 @@ export const Combobox = React.forwardRef(function Combobox(
   let listboxId = id ? makeId("listbox", id) : "listbox";
 
   let isControlledRef = React.useRef<boolean>(false);
-  let setIsControlled = (val: boolean) => (isControlledRef.current = val);
 
   let context: InternalComboboxContextValue = {
     ariaLabel,
@@ -331,8 +330,7 @@ export const Combobox = React.forwardRef(function Combobox(
     popoverRef,
     state,
     transition,
-    isControlled: isControlledRef.current,
-    setIsControlled,
+    isControlledRef,
   };
 
   useCheckStyles("combobox");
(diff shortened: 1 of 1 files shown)


packages/combobox/src/index.tsx
```diff
@@ -312,6 +312,8 @@ export const Combobox = React.forwardRef(function Combobox(
   let id = useId(props.id);
   let listboxId = id ? makeId("listbox", id) : "listbox";
 
+  let [isControlled, setIsControlled] = React.useState(false);
```
chaance, line 315: Is there a reason we need new state for this? Could we not store this as a ref instead as we do lower in the tree? I don't believe we ever need it for rendering.

```ts
let isControlledRef = React.useRef(false);

// in ComboboxInput:
let isControlled = controlledValue != null;

React.useEffect(() => {
  isControlledRef.current = isControlled;
}, [isControlled]);

// In ComboboxOption:
let { isControlledRef } = React.useContext(ComboboxContext);
let handleClick = () => {
  onSelect && onSelect(value);
  transition(SELECT_WITH_CLICK, { value, isControlled: isControlledRef.current });
};
```

arackaf: @chaance that's exactly what I did originally, but decided to go with state, since setting that ref would not trigger a re-render, so I figured state would be more safe. 

It would almost certainly work either way though. If you think a ref would be better I could change to that.

arackaf: @chaance ok I've got it ref based now. I'm putting just the boolean in context, as opposed to the ref itself. If you'd like me to remove the updater function, and just store the ref object, lemme know and I will.

chaance (approved):

chaance:
Thanks for the changes, looks good 👍

Issue #394, opened 2019-11-28
[Alert] Screen reader sees and reads both alerts when full page read is requested


mareksuscak:
### Current Behavior

Right now, the screen reader sees both the visible alert and the live region due to the fact that we're [essentially](https://github.com/reach/reach-ui/blob/18281e8be79e2cf2399408b835e5f0879f657592/packages/alert/src/index.js#L110-L122) [duplicating](https://github.com/reach/reach-ui/blob/18281e8be79e2cf2399408b835e5f0879f657592/packages/alert/src/index.js#L48-L53) `Alert`'s children and none of them is hidden from the screen reader by applying a `role="none"` or `aria-hidden="true"`.

See the example below.

### Expected behavior

When the full page read is requested (on iOS use two fingers and swipe from top to bottom)

### Reproducible example

[CodeSandbox Example](https://codesandbox.io/s/reach-ui-template-rceon)

### Suggested solution(s)

One of the following:

1. Do not render children twice (see WAI-ARIA authoring practices [example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/alert/alert.html)).
2. Hide the visible content from the screen reader (not quite sure if there's a use case that requires both to be visible) with `role="none"` or `aria-hidden="true"`

### Your environment

| Software                 | Name/Version(s) |
| ------------------------ | --------------- |
| Package Name | **@reach/alert**
| React                    | 16.x.x
| Browser                  | Safari on iOS
| Assistive tech           | VoiceOver on iOS
| Node                     | 10.17.0
| npm/yarn                 | 1.19.2
| Operating System        |  iOS 13

mareksuscak:
Can we reopen please? This is definitely still an issue.

hchiam:
Interesting - I'll take a look.

hchiam:
I can reproduce the bug with VoiceOver on macOS: the alert is announced twice when I go to the provided example https://rceon.csb.app

hchiam:
As per [the MDN doc that mentions `aria-hidden` and `role="none"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden), adding either `aria-hidden` or `role="none"` to the `aria-live` region doesn't seem to do what we'd want: 
[...]

PR #906, opened 2022-02-07
[Alert]: Make the Alert itself aria-live to dedupe screen reader announcement

hchiam:
This pull request fixes bug #394 in an existing package ([Alert](https://github.com/reach/reach-ui/tree/develop/packages/alert)):

- [Visually hidden or not, content that dynamically changes can be used as an `aria-live` region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
- I tested http://localhost:9001/iframe.html?id=alert--basic-ts-ts&viewMode=story with VoiceOver on macOS (I was able to repro the bug before my changes, and after my changes I didn't hear the duplicated announcement). Also tested on Firefox and on Chrome. `control`+`option`+`a` to read whole page

### DIFF
--- a/packages/alert/src/index.tsx
+++ b/packages/alert/src/index.tsx
@@ -7,7 +7,7 @@
  * readers, and in some operating systems, they may trigger an alert sound.
  *
  * The approach here is to allow developers to render a visual <Alert> and then
- * we mirror that to a couple of aria-live regions behind the scenes. This way,
+ * we treat it as an aria-live region behind the scenes. This way,
  * most of the time, developers don't have to think about visual vs. aria
  * alerts.
  *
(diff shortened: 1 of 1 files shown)


hchiam (commented):
Ready for review.

TimKolberger:
I think this might solve #916, because it removes the usage of `ReactDOM.render`.

chaance:
I think the better solution (for now) is to add `aria-hidden` to the visible container as suggested in #394. I think the reason it didn't work in your testing is because cloning the child component clones that property to the announcement node, which we obviously don't want. A little refactoring to ensure it's only added to the visible container should do the trick.

Issue #925, opened 2022-05-20
Support React 18


tmthan:
React pulished version 18.
Please upgrade library to support React 18.

ryparker:
Duplicate of https://github.com/reach/reach-ui/issues/916

PR #939, opened 2022-07-01
Updated portal and utils to accept React 18 in package.json

PatrikTheDev:
This pull request simply updates the React dependency version for portal and utils (I haven't tested other packages on React 18 so I'm not comfortable updating those yet).


chaance:
There is code in the `utils` package that isn't ready for React 18. See https://github.com/reach/reach-ui/issues/925 for context. We'll get all of the peer deps updated when I'm confident in React 18 support throughout the repo!

PR #941, opened 2022-07-05
Repo overhaul + housekeeping

chaance:
This PR overhauls much of the core structure of the repo as well as many of its underlying tooling. The overall task is a WIP, but to start the main points are:

- Ditching Yarn 1/Lerna/Preconstruct for pnpm + Turbo
- Ditching jest for Vitest
- Ditching Babel for Tsup
- Moving Storybook dependencies into a separate workspace for clarity
- Moves all examples into Storybook workspace and adds symlinks in package directories


brennarvo:
@chaance Thanks for adding an option to test React 18.

Only issue I'm seeing is that vitest can't resolve React when running `pnpm test:react-18` or `pnpm test:react-16`. 

Screen Shot 2022-07-13 at 9 18 41 PM


I thought this was an issue with the resolve alias, so I updated the Vitest resolve alias config to match the [`AliasOptions` interface](https://github.com/vitejs/vite/blob/cda5442fc5eae9a571536fb2fcea9d9f16c28365/packages/vite/types/alias.d.ts#L32) with an absolute path resolution to my node_modules directory.

After switching the config, the same `TypeError` was being thrown. I'm thinking the [@vitejs/plugin-react](https://github.com/vitejs/vite/tree/main/packages/plugin-react#vitejsplugin-react-) would help out here, but that also needs Vite itself as a dependency.

Also one thing I noticed in the docs:
> [Vitest requires Vite >=v2.7.10](https://vitest.dev/guide/#trying-vitest-online)

How were you able to get Vitest running without Vite?

Issue #916, opened 2022-04-04
Compatibility with React 18?


nemdo:
Components are not compatible with React 18 as it can be seen in package-lock.json:
```
"peerDependencies": {
    "react": "^16.8.0 || 17.x",
    "react-dom": "^16.8.0 || 17.x"
}
```
Eslint throws warnings, but shows 0 vulnerabilities and I haven't had any issue yet.

Is it safe to use Reach UI with React 18? When will you update the libraries? Thank you in advance!

johnnyluu:
I'm finding that Listbox doesn't work (no response after clicking on the trigger) in my new project using React 18, but the same setup works in React 17.

chaance:
I don't support React 18 just yet. There are a number of things that won't play well with concurrent rendering or the new StrictMode behavior. I'm going to start going through all of that this weekend and take stock of what all needs to be done to get there.

tajo:
It would be nice to just relax the `peerDeps` versions of React first. It seems to be mostly working but some package managers like pnpm are more strict and just error out if there's peerDeps mismatch.

jeanpan:
> I don't support React 18 just yet. There are a number of things that won't play well with concurrent rendering or the new StrictMode behavior. I'm going to start going through all of that this weekend and take stock of what all needs to be done to get there.

@chaance, would you mind to list what all needs to be done here?

chaance:
@jeanpan Main thing is checking everything out in `StrictMode` now that React double executes effects, which will likely spot a few bugs we need to fix. In a few places we are tracking renders in the effect (eg, `usePrevious`), and that's gonna be a problem. Another thing is to ensure that we aren't improperly accessing refs during render to prevent concurrent rendering bugs.

There may be some other things, but running the test suite w/ React 18 and checking the storybook examples would be a good way to find out for sure.

hood:
Any news on this?

brennarvo:
I tried running the new `test:react-18` script inside the repo to see if the tests pass with React 18, but was getting path resolution errors: https://github.com/reach/reach-ui/pull/941#issuecomment-1183910011

chaance:
`test:react-16` and `test:react-18` haven't been properly set up yet. I'll respond here when I have everything working.

nemdo:
4 months have passed since the issue was opened and among the 30 dependencies in my project, @reach/dialog is the only one that is not compatible with react 18.x.x. When will you release the new version approximately?

[...]

PR #953, opened 2022-08-05
Add peer dependency for React 18 in `visually-hidden` package

cassidoo:
Hey team, right now if you install this package with a fresh React 18 project, you get a `ERESOLVE unable to resolve dependency tree` and I made this small change so that not everyone needs to do the whole `--legacy-peer-deps` thing.

I can update it to `>=16.8` instead of adding the `||` if you'd prefer, too, and add other packages as well, but I started small in case there might be a reason to not do this.

This pull request:
- Adds additional features/functionality to an existing package

### DIFF
--- a/packages/visually-hidden/package.json
+++ b/packages/visually-hidden/package.json
@@ -21,8 +21,8 @@
     "react-dom": "^17.0.2"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || 17.x",
-    "react-dom": "^16.8.0 || 17.x"
+    "react": "^16.8.0 || 17.x || 18.x",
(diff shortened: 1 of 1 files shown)


bradwestfall:
I get this a lot for the other packages too, not just visually-hidden. This might be more of a @chaance thing though, I'm not so much involved.

cassidoo:
Good to know @bradwestfall! I figured Chance might be the person to ping on this but wasn't sure 😅

chaance:
For reference there is still a lot of testing needed for React 18 in general ([tracking here](https://github.com/reach/reach-ui/issues/916)). This one is safe to update though so I'll merge 👍

PR #959, opened 2022-08-24
Setup tests run for react@16 and react@18

dartess:
Hi. 
It wasn't easy, but I think I set everything up to run scripts `test:react-16` and `test:react-18`.


chaance:
Super helpful, thank you!

PR #960, opened 2022-08-25
Add tests for '@reach/utils' and compatibility check with react 18

dartess:


packages/utils/src/use-focus-change.ts
```diff
@@ -22,20 +22,24 @@ export function useFocusChange(
...
 		function onChange(event: FocusEvent) {
-			if (lastActiveElement.current !== ownerDocument.activeElement) {
-				handleChange(
-					ownerDocument.activeElement,
-					lastActiveElement.current,
-					event
-				);
-				lastActiveElement.current = ownerDocument.activeElement;
+			if (
+				when === "focus" &&
```
dartess, line 26: If do not check it and try use this hook with `blur`, handleChange will never be called, because `ownerDocument.activeElement` will be always `ownerDocument.body`.

packages/utils/src/use-focus-change.ts
```diff
@@ -22,20 +22,24 @@ export function useFocusChange(
...
+				event
+			);
+			lastActiveElement.current = ownerDocument.activeElement;
 		}
 
 		ownerDocument.addEventListener(when, onChange, true);
 
 		return () => {
-			ownerDocument.removeEventListener(when, onChange);
+			ownerDocument.removeEventListener(when, onChange, true);
```
dartess, line 42: removing the handler didn't work before.
demo: https://codepen.io/dartess/pen/rNdgWEj

PR #961, opened 2022-08-28
Fix `auto-id` build for backwards compatibility with React <18

### LINKED ISSUE #921: 0.17.0 does not work with React 17
## 🐛 Bug report

Since useId is not available on React17 it breaks with reach@0.17.0

In auto-id package https://github.com/reach/reach-ui/blob/develop/packages/auto-id/src/index.tsx,  

```import * as React from "react";``` 

is compiled as 

```import { useId as useId$1, useState, useEffect } from 'react'```

 in the esm module.

This leads to the following error 

image

| Reach Package    |  reach/alert-dialog   |    0.17.0     |
| React            |     React    |   17.0.2      |
| React-scripts           |     react-scripts    |  5.0.0     |
| npm/yarn         |     yarn     |   1.22.18      |
| Operating System |    MacOs     |     12.3.1    |

### PR DESCRIPTION
**Pull request by emmenko**
## Summary

Fixes #921

TL;DR: the bundle of `@reach/auto-id` contains an import of `useId` from `react`. For versions of React <18 this import does not exist, causing an error when this package is used.

image

This can also be seen in the Storybook playground build output (although Storybook shows a warning)

image

To fix that, I suggest to remove that code and change the implementation once React >=18 is supported.

I also fixed another import issue shown by Storybook

image

---

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ node_modules
 .env
 coverage
 
+# storybook
+/playground/storybook-static
+
 # website
 /website-deploy-key
 /website-deploy-key.pub
--- a/packages/auto-id/src/auto-id.ts
+++ b/packages/auto-id/src/auto-id.ts
@@ -87,15 +87,6 @@ function useId(
 function useId(): string | undefined;
 
 function useId(providedId?: number | string | undefined | null) {
-	// TODO: Remove error flag when updating internal deps to React 18. None of
-	// our tricks will play well with concurrent rendering anyway.
-	// @ts-expect-error
-	if (typeof React.useId === "function") {
-		// @ts-expect-error
-		let id = React.useId(providedId);
-		return providedId != null ? providedId : id;
-	}
-
 	// If this instance isn't part of the initial render, we don't have to do the
 	// double render/patch-up dance. We can just generate the ID and return it.
 	let initialId = providedId ?? (serverHandoffComplete ? genId() : null);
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -12,6 +12,9 @@ node_modules
 .env
 coverage
 
+# storybook
+/playground/storybook-static
```
emmenko, line 16: Artifact of building Storybook locally.

packages/auto-id/src/auto-id.ts
```diff
@@ -87,15 +87,6 @@ function useId(
...
 
 function useId(providedId?: number | string | undefined | null) {
-	// TODO: Remove error flag when updating internal deps to React 18. None of
-	// our tricks will play well with concurrent rendering anyway.
-	// @ts-expect-error
-	if (typeof React.useId === "function") {
-		// @ts-expect-error
-		let id = React.useId(providedId);
-		return providedId != null ? providedId : id;
-	}
```
emmenko, line 97: This is what is causing the problem. I suggest to simply remove this once only React >=18 is supported.

playground/stories/window-size/basic.example.js
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import WindowSize from "@reach/window-size";
+import { WindowSize } from "@reach/window-size";
```
emmenko, line 2: The import was wrong

packages/auto-id/src/auto-id.ts
zkuzmic: Is there no way to rewrite this to keep your build process from showing a warning? I think that would be preferable.

PR #970, opened 2022-09-14
bug:menu-button - fix duplicated menu-button id with multiple instances

jeanpan:
This is to fix the issue when there is no `id` set specifically in `menu-button`, the `id` is `menu-button--menu`. It causes duplicated `id` if there are multiple `menu-button` instances on the page. 

Before the fix

Screen Shot 2022-09-14 at 15 55 29

After the fix

Screen Shot 2022-09-14 at 16 21 37

**Root cause**
It's caused by the `triggerId` isn't updated since there is no dispatch action to update the state when the `auto-id` becomes valid. 

---

- [ ] Use a meaningful title for the pull request. Include the name of the package modified.

This pull request:

- [ ] Creates a new package
- [ ] Other

If creating a new package:

  - [ ] `examples` directory


jeanpan:
@chaance, would you mind to review this change? I'd like to have author to review the change to make sure I didn't miss anything. Thank you.

chaance (approved):

PR #980, opened 2022-10-17
Add peer dependency for React 18 in combobox package

drDre-code:
Hey guys, right now if you install this package with a fresh React 18 project, you get a `ERESOLVE unable to resolve dependency` tree and I made this small change so that not everyone needs to do the whole `--legacy-peer-deps` thing.


ben-heimberg:
is it gonna happen?

